PR #41, opened 2022-02-10
Fix docker compose example link

mmattel:
This fixes the docker compose example link used by wget

Fixes: https://github.com/owncloud/docs/issues/4692 (bad URL in docker compose instructions)

Backport to 10.9 and 10.8


phil-davis (approved):
I wonder how many other mentions like this  are now out-of-date since the content moved to docs-server repo?

mmattel:
> I wonder how many other mentions like this are now out-of-date since the content moved to docs-server repo?

There will be some, but I strongly believe you can count them on one hand

PR #50, opened 2022-02-11
Remove mistake entry about incompatibility with storage encryption

pmaier1:
@mmattel please take care for back (forward?) ports.


mmattel:
Backport to master only, 10.8 does not have this part

PR #75, opened 2022-02-17
Update ubuntu_20_04.adoc

voroyam:
- removed unnecessary variables
- moved notes to the prerequisites section
- moved apache and database to preparation section
- added smb client php module installation instruction

Reason for removing the variables: if someone wants to make a copy of this quick guide in raw form and execute it offline - he would fail because of the variables and include scripts.

Most of them are irrelevant, for example: we are ubuntu apache exclusive, so the default web server user is always www-data.


modules/admin_manual/pages/installation/quick_guides/ubuntu_20_04.adoc
```diff
@@ -65,9 +67,23 @@ apt install -y \
   php-apcu php-redis php-ldap 
 ----
 
-Note : php 7.4 is the default version installable with Ubuntu 20.04
+=== Install smb client php module
 
-See the important note on using the correct xref:installation/manual_installation/manual_installation_apache.adoc#multi-processing-module-mpm[Multi-Processing Module (MPM)].
+[source,console]
+----
+apt-get install -y libsmbclient-dev php-dev
```
enbrnz, line 74: I think php-pear also was another requirement

modules/admin_manual/pages/installation/quick_guides/ubuntu_20_04.adoc
```diff
@@ -1,57 +1,59 @@
 = Install ownCloud on Ubuntu 20.04
 :toc: right
-:logrotate-url: https://linux.die.net/man/8/logrotate
-:webserver-user: www-data
-:webserver-group: www-data
-:install-directory: /var/www/owncloud
-:page-aliases: installation/quick_guides/ubuntu_18_04.adoc
```
enbrnz, line 7: ^ Not sure if it is wanted to remove all of these?

enbrnz (changes requested):

modules/admin_manual/pages/installation/quick_guides/ubuntu_20_04.adoc
voroyam: Those are variable values which are called later in the document. But with no variables being in the document anymore - you don't need the variables being set at the top.

Last line was just weird to me. This is a 20.04 guild, why is it a relay for a 18.04 document? No one who was looking for 18.04 is going to be happy to find a 20.04 document.

modules/admin_manual/pages/installation/quick_guides/ubuntu_20_04.adoc
voroyam: Yes. will add.

mmattel:
I will summarize my comments here without doing it line by line from top down.

* Keep the page alias, it will point external links to 20.04 without creating a 404. It was made by intention
* OK to hard-code www-data
* Concatenate line 16 and 17, no need to have a linebreak which will be rendered anyways as single line
* If you remove an include, see line 96, also remove the file from the source folder (if it is nowhere else used which I believe is the case here), and if it was the last file in that folder, remove as many parent folders that have no files present. Avoid an orphand file
* Line 126/129, this is not ok, undo the changes. We DO NOT hardcode a release name but use the attribute as it was there. We recently had a name change on that file. A single change in the attribute changes at all the locations where used
* Same for  L198/283, use the attribute

When changed, I will review again and approve/merge/backport it

voroyam:
> I will summarize my comments here without doing it line by line from top down.
> 
> * Keep the page alias, it will point external links to 20.04 without creating a 404. It was made by intention

ok

> * OK to hard-code www-data

ok

> * Concatenate line 16 and 17, no need to have a linebreak which will be rendered anyways as single line

ok

> * If you remove an include, see line 96, also remove the file from the source folder (if it is nowhere else used which I believe is the case here), and if it was the last file in that folder, remove as many parent folders that have no files present. Avoid an orphand file

Yeah, I don't know where that stuff is anymore. Need someone from docs team to point me to it so I can remove it.

> * Line 126/129, this is not ok, undo the changes. We DO NOT hardcode a release name but use the attribute as it was there. We recently had a name change on that file. A single change in the attribute changes at all the locations where used

If I take responsibility for that guide, because I am always using it, and you don't because you like the detailed guilde - not sure what's the harm. I put the name of the release which is latest, which will have always the latest packages, it's not like I put a date in there that will be incorrect after the next release.

> * Same for  L198/283, use the attribute

No idea what that is. can't find it. Sorry.
> 
> When changed, I will review again and approve/merge/backport it

Can you point me to that stuff? especially the last 2, I can't find it based on the line coordinates.

mmattel:
Calm down, you have noticed that the PR, looks great 👍 reopen, fix the last item and we are done.

voroyam:
I don't work for you.

modules/admin_manual/pages/installation/quick_guides/ubuntu_20_04.adoc
```diff
@@ -195,27 +232,44 @@ occ config:system:set \
    --value '\OC\Memcache\Redis'
 occ config:system:set \
    redis \
-   --value '{"host": "127.0.0.1", "port": "{std-port-redis}"}' \
+   --value '{"host": "127.0.0.1", "port": "6379"}' \
```
EParzefall, line 235:
```suggestion
   --value '{"host": "127.0.0.1", "port": "{std-port-redis}"}' \
```

EParzefall (approved):
Should be everything now. Orphaned include file will be deleted in a separate PR after checking if it is included anywhere else.

PR #102, opened 2022-02-19
[10.9] [PR 75] Update ubuntu_20_04.adoc

EParzefall:
Backport of PR #75


mmattel (approved):
Have you checked that the removed include file is not orphand now?
In case you need to create another PR removing it...

EParzefall:
> Have you checked that the removed include file is not orphand now? In case you need to create another PR removing it...

That's the plan. There's another file in the 18.04 examples which probably needs to go as well.

PR #104, opened 2022-02-19
deleting orphans

EParzefall:
Deleting orphans left behind after removal of Ubuntu 18.04 chapter.

Backport to 10.9 ~and 10.8~ --> **10.9 only** as U18.04 actually exists in 10.8 (updated comment by mm)


mmattel:
Note, have you checked that the files are not used in `master` anywhere else? Just for security that we do not remove files that have a remaining reference...

mmattel (approved):

EParzefall:
> Note, have you checked that the files are not used in `master` anywhere else? Just for security that we do not remove files that have a remaining reference...

Both files are not referenced anywhere in docs-server master.

Issue #109, opened 2021-08-27
How to configure pollinterval capability to control desktop and iOS app


michaelstingl:
## WHAT Needs to be Documented?

How to configure pollinterval capability

## WHERE Does This Need To Be Documented (Link)?

https://doc.owncloud.com/server/admin_manual/

## WHY Should This Change Be Made?

It's currently not documented

## (Optional) What Type Of Content Change Is This? 
- [x] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

Related:
- https://github.com/owncloud/client/pull/8777
- https://github.com/owncloud/ios-app/pull/1021

phil-davis:
I have documented by:
1) adjusting the existing developer docs (PR #110 )
2) adding to `config.sample.php` in core PR https://github.com/owncloud/core/pull/39143 - that will come to here in a config-to-docs run

Is another descriptive section needed for this somewhere in the admin docs?  Or?

mmattel:
Btw, the desktop and ios is mentioned, what about android? Is it implemented there?
Adding @jesmrec maybe he knows more

jesmrec:
Polling interval is not developed in Android yet.

JRundfeldt:
The Docs need the info when the clients get the new value - immediatly? at the next request? at login? after reboot of a client system?

phil-davis:
The server provides the value whenever it is asked. The docs of each client would need to be enhanced to say when the client checks/rechecks that capability (or, for example, on Android add documentation to say that the setting is ignored)

jnweiger:
In 10.10.0 rc2 the capabilities say pollinterval 30000

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

The requested documentation for the `pollinterval` capability now exists, so this is closed as resolved.

**Admin manual** — the `pollinterval` config option is documented in `config_sample_php_parameters.adoc` under the section *"Define the suggested poll interval for clients"*:

> Specifies how often clients should poll the server for changes. The value is in milliseconds. The value is not enforced. Clients may use this value to decide how frequently to check the server for changes.

```php
'pollinterval' => 30000,
[...]

PR #110, opened 2022-02-22
Adjust pollinterval in docs

phil-davis:
Part of issue #109 

Note: after core PR https://github.com/owncloud/core/pull/39143 is merged, a config-to-docs run is needed. `pollinterval` was not previously mentioned in config.sample.php and is added in that PR.

No backport needed - this will be in the next oC10 release (might be numbered 10.10.0)

### DIFF
--- a/modules/developer_manual/examples/core/apis/ocs-capabilities/list-capabilities-response.json
+++ b/modules/developer_manual/examples/core/apis/ocs-capabilities/list-capabilities-response.json
@@ -76,7 +76,7 @@
                   "maintenance" : "false",
                   "version" : "10.0.3.3"
                },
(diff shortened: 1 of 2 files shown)


mmattel:
pls no direct changes in any of the both config.sample files as this may confuse a later ctd run. the original change must happen in core (merged) and can further be transferred with a ctd run.

phil-davis:
> pls no direct changes in any of the both config.sample files as this may confuse a later ctd run. the original change must happen in core (merged) and can further be transferred with a ctd run.

this PR does not have any cofig.sample.php change

EParzefall (approved):
LGTM

PR #117, opened 2022-02-24
Enable encryption via occ is incomplete and leads to deprecated user-key mode

jnweiger:
One step is mentioned, but it takes actually 3 steps.
Without the third one, we run into https://github.com/owncloud/encryption/issues/327#issuecomment-1049845296


EParzefall (approved):
LGTM and thanks!

EParzefall:
Backports to 10.9 and 10.8, I assume?

PR #128, opened 2022-03-01
Source language fix for harden server

mmattel:
Missed fixes for the harden server document

Note related, but added a qnap text fix as it needs the same backporting

Backport to 10.9 and 10.8


EParzefall (approved):
LGTM

PR #133, opened 2022-03-02
Updates to innodb tunning instructions,

crrodriguez:
innodb_read_io_threads and innodb_write_io_thread defaults to 4, no need to change the defaulr

innodb_flush_log_at_trx_commit=1 is the default as well

innodb_thread_concurrency does nothing since  mariadb 10.5.5
innodb_file_per_table=ON same

innodb_large_prefix = 1 is a no-op, has not done anything since maria db 10.3


mmattel (approved):

mmattel:
Backporting to 10.9 and 10.8

PR #157, opened 2022-03-05
Fix some broken links found after migrating the user manual

mmattel:
This PR fixes some broken internal links which were found after a full build post the user manual move with `htmltest`. Note that this are not broken links because of the migration.

Backport to 10.9 and 10.8
(Note that we have to check the 10.8 backport for stuff not belonging to there...)


EParzefall (approved):
LGTM

PR #168, opened 2022-03-07
Update ubuntu_20_04.adoc

jnweiger:
not sure, why we introduce a vairable and then use it once. 
Either use it everywhere, or don't introduce.


mmattel:
> not sure, why we introduce a variable and then use it once. Either use it everywhere, or don't introduce.

Pls note that the quick guide is maintained by the support team, the docs team only takes care about keeping it inline with general docs rules.

mmattel (approved):

PR #185, opened 2022-03-10
expand install hints for postgresql (a bit)

jnweiger:
postgresql install instructions facelift.
- add missing dependency package (for ubuntu)
- do not mess with postgres user - (that is a super administrator with all permissions.)
- instead add an owncloud user who only has control over the owncloud database.

(With `occ maintenance:install` the `--database pgsql` parameter choses postgresql  -- not sure, where this information would fit...)

I assume, postgresql 12 and 13, 14 work just fine with owncloud.
An official statement about supported versions should come from the nepal QA team, they start testing now.

### DIFF
--- a/modules/admin_manual/pages/installation/manual_installation/manual_installation_db.adoc
+++ b/modules/admin_manual/pages/installation/manual_installation/manual_installation_db.adoc
@@ -43,19 +43,17 @@ To install postgres, use the following command (or that of your preferred packag
 
 [source,bash]
 ----
-sudo apt-get install postgresql
+sudo apt-get install postgresql php-pgsql
 ----
 
-In order to allow ownCloud access to the database, create a known password for the default user, `postgres`, which was added when the database was installed.
+In order to allow ownCloud access to the database, create a an ownlcoud user who owns the owncloud database.
 
 [source,bash]
 ----
-sudo -i -u postgres psql
-postgres=# \password
+sudo -u postgres -c "createuser -e -P owncloud" 
 Enter new password:
 Enter it again:
-postgres=# \q
-exit
+sudo -u postgres -c "createdb -e -O owncloud owncloud
 ----
 
 == Oracle 11g
--- a/modules/admin_manual/pages/installation/manual_installation/manual_installation_db.adoc
+++ b/modules/admin_manual/pages/installation/manual_installation/manual_installation_db.adoc
@@ -46,14 +46,25 @@ To install postgres, use the following command (or that of your preferred packag
 sudo apt-get install postgresql php-pgsql
 ----
 
-In order to allow ownCloud access to the database, create an ownlcoud user who owns the owncloud database.
+In order to allow ownCloud access to the database, create an `ownlcoud` user who owns the `owncloud` database. The user and the database name can be any name that fits your needs.
 
+Create the `owncloud` PostgreSQL user account. Note that the default admin account post installing PostgreSQL is `postgres`. If you have created another postgres admin user and disabled the default one, use the new one instead.
 [source,bash]
 ----
-sudo -u postgres -c "createuser -e -P owncloud" 
+sudo -u postgres -c "createuser -e -P owncloud"
+----
+
+Define a password for the `owncloud` user. Note to remember this password to access the database later on:
+[source,plaintext]
+----
 Enter new password:
 Enter it again:
-sudo -u postgres -c "createdb -e -O owncloud owncloud
+----
+
+Create the `owncloud` database owned by the `owncloud` user:
+[source,bash]
+----
+sudo -u postgres -c "createdb -e -O owncloud owncloud"
 ----
 
 == Oracle 11g
--- a/modules/admin_manual/pages/installation/manual_installation/manual_installation_db.adoc
+++ b/modules/admin_manual/pages/installation/manual_installation/manual_installation_db.adoc
@@ -46,7 +46,7 @@ To install postgres, use the following command (or that of your preferred packag
 sudo apt-get install postgresql php-pgsql
 ----
 
-In order to allow ownCloud access to the database, create a an ownlcoud user who owns the owncloud database.
+In order to allow ownCloud access to the database, create an ownlcoud user who owns the owncloud database.
 
 [source,bash]
 ----


modules/admin_manual/pages/installation/manual_installation/manual_installation_db.adoc
```diff
@@ -43,19 +43,17 @@ To install postgres, use the following command (or that of your preferred packag
 
 [source,bash]
 ----
-sudo apt-get install postgresql
+sudo apt-get install postgresql php-pgsql
 ----
 
-In order to allow ownCloud access to the database, create a known password for the default user, `postgres`, which was added when the database was installed.
+In order to allow ownCloud access to the database, create a an ownlcoud user who owns the owncloud database.
```
phil-davis, line 49:
```suggestion
In order to allow ownCloud access to the database, create an ownlcoud user who owns the owncloud database.
```

mmattel:
I have improved the stuff a bit by adding more explanation texts, splitting the commands for proper copy/paste and added a missing end quote in one command.

Question: imho we should fix the [system requirements](https://doc.owncloud.com/server/next/admin_manual/installation/system_requirements.html) documentation (if the version technically works with oC)
change: `PostgreSQL 9 and 10` --> `PostgreSQL 9.4 and above, note that Ubuntu 20.04 ships with version 12.2.14`

@phil-davis any objections?
@jnweiger can you have a look on the changes made?

mmattel:
> the `--database pgsql` parameter choses `postgresql`

what is meant by that?

phil-davis:
> Question: imho we should fix the [system requirements](https://doc.owncloud.com/server/next/admin_manual/installation/system_requirements.html) documentation (if the version technically works with oC)

We are currently running tests with pgsql 11, 12, 13 (and we will try 14...). Issue https://github.com/owncloud/enterprise/issues/5064

I created issue #186 to remind us to update the docs when the testing results are known.

phil-davis (approved):
LGTM

jnweiger:
> > the `--database pgsql` parameter choses `postgresql`
> 
> what is meant by that?

When installing with occ maintenance:install (not sure, if that is documented at all, if not, ignore), **and** php bindings for multiple databases are present in the system, then the option `--database` can be used to choose between sqlite, mysql, and postgres. (and maybe oracle, never tried that)

The values to use with that option for sqlite and mysql are the obvious `sqlite` and `mysql` -- but for postgesql you have to use `pgsql`. The online help for `occ maintenance:install` also does not list the possible values. So users currently have to guess.

That is what I meant to say... 

----

Regarding your additions:
-  "can be any name that fits your needs" -- that is mostly true, but not particularly suited here, as postgres is very picky with special characters. We don't want to detail the rules here.
- "default admin account post installing" and "If .. disabled the default one, use the new one" -- all that is also true, but probably more confusing than helpful. The english languange level is complex. Kowledge about postgres security modes (and the pg_hba.conf jungle) is prerequisite.

I'd just give some rationale for the choices we made, without discussing options:

```We call our database "owncloud" and also create an "owncloud" user for it. To authorize the createdb and createuser commands, we use the Linux postgres user account, that was implicitly added when installing the postgresql package.```
 
In addition to the password, we set for the new owncloud account, both `sudo` and `createuser` could ask for passwords while connecting, if the system is set up in a non-default way. ...  maybe we should split the command in two to help the reader get everything sorted?

**Comment by jnweiger** [comment]
When multiple php database bindings are available, and the web based installation wizard is used, 
the user normally only sees a warning that a bad choice will be used. 
[image](https://user-images.githubusercontent.com/1108546/157952591-fdb955c2-6f97-42fc-a39d-37698d6c3554.png)

----

In reality, the "bad choice" is not yet made. The little downward triangle opens a choice, where the available databases are spelled out nicely:
image

**Comment by jnweiger** [comment]
When multiple php database bindings are available, and the web based installation wizard is used, 
the user normally only sees a warning that a bad choice will be used. 
[image](https://user-images.githubusercontent.com/1108546/157952591-fdb955c2-6f97-42fc-a39d-37698d6c3554.png)

----

In reality, the "bad choice" is not yet made. The little downward triangle opens a choice, where the available databases are spelled out nicely:
image

**Comment by mmattel** [comment]
> When installing with occ maintenance:install (not sure, if that is documented at all, if not, ignore), **and** php bindings for multiple databases are present in the system, then the option `--database` can be used to choose between sqlite, mysql, and postgres. (and maybe oracle, never tried that)

Please help, I cant find a `occ maintenance:install` command. I am running oC 10.9.1 and in the documentation and when checking
```
sudo -uwww-data ./occ maintenance

  Command "maintenance" is not defined.

  Did you mean one of these?
      maintenance:data-fingerprint
      maintenance:mimetype:update-db
      maintenance:mimetype:update-js
      maintenance:mode
      maintenance:repair
      maintenance:singleuser
      maintenance:update:htaccess
```
`:install` is missing...
So how can one run an occ command that must fail because of missing accessability(see quick and full install guide)? Note that the code is present in core and in my running instance see: https://github.com/owncloud/core/blob/master/core/Command/Maintenance/Install.php
Is there some code that prevents showing it when the db has been setup already?

phil-davis:
https://github.com/owncloud/core/blob/master/core/register_command.php#L213

The maintenance:install command is only registered if the system is not installed. Locally you can set "installed" to `false` and then the command will be available.

mmattel:
> The maintenance:install command is only registered if the system is not installed

Thanks for clarification. I have to add the command in the description and a note about this behaviour.

mmattel:
@jnweiger question regarding the screenshot below:
image
Does this change if an enterprise licence is added before doing the setup so that oracle can be selected with the wizard?

jnweiger:
No, I don't think so. Which choices appear there, depends on the installed php-database modules.

mmattel:
> No, I don't think so. Which choices appear there, depends on the installed php-database modules.

I just read that you have install the php-oci8 extension via [pecl](https://pecl.php.net/package/oci8).
should we add to the wizard section that only those databases appear where the db php component is installed?

jnweiger:
Dangerzone ahead:
pecl/oci8 requires PHP (version >= 8.1.0), installed version is 7.4.3

I am not ready to speak about oracle at all. Happy to have mastered PostgeSQL by now.

mmattel:
> Dangerzone ahead:

Bomb disarmed, see https://github.com/owncloud/docs-server/pull/190 (Define the correct oci8 release for pecl because of PHP7.x)

jnweiger:
Tse, "this guide" -- who knows what people would try there... :sparkle:  :bomb:

PR #191, opened 2022-03-14
link to "this guide" added.

jnweiger:
Hard to guess, what is meant with "this guide", when there is no link. 
I found one. But I don't think it qualifies as thorough. It is awkward, it uses the alien tool.

*untested*(tm)


mmattel (approved):

mmattel:
Backport to 10.9 and 10.8

PR #217, opened 2022-03-28
add secure passwords

voroyam:
- added password generators for admin and db passwords
- updated the db creation and oC installation with variables containing passwords
- added an output of the password variables at the end
- added a link to the ownCloud instance at the end

Backport to 10.9 and 10.8


mmattel (approved):
LGTM, though not tested

PR #224, opened 2022-03-30
Remove Fedora

enbrnz:
Fedora has a much faster lifecycle (6 months?) and is, because of this in my opinion not suitable for server applications. Additionally, I don't know of any customers actually using it.

Edit: additionally, Fedora 32 & 33 are both EoL: https://docs.fedoraproject.org/en-US/releases/eol/


mmattel:
Closing, superseded by #227 (Update system_requirements.adoc) which includes that change.

PR #225, opened 2022-03-30
remove variable

voroyam:
the occ wrapper script is described in the installation guide and is being used on all deployments and docker installations.

there is no need for the sudo part.


mmattel:
Sorry to say, but there is no agreement to this change and it will not be approved nor merged.
occ commands are and have to always be written generally in all documents where used (with only a very view exceptions) and not specifically for the situation as created in the owncloud quick install guide.

@EParzefall fyi

PR #226, opened 2022-03-30
Updates to hardening guide

crrodriguez:
-  brcrypt does not limit the password to 72 chars due to the reasons stated, but due to the limits of the underlying cipher it uses (blowfish) which only fully mixes up to 448 bits (56 chars) of the key.. but most implementations place an absolute limit of 576 bit  (72 chars) 

- Remove the section about /dev/urandom, it is clumsily worded and outdated, PHP now only uises /dev/urandom as a fallback and raises errors or exceptions if it cannot be opened. rfc4086 is not relevant to owncloud, meither to PHP which just consumes the OS entropy sources.

Backport to 10.9 and 10.8


mmattel (approved):

mmattel:
Thanks for taking care on 👍

PR #241, opened 2022-04-11
put the user-key is legacy in the title

voroyam:
for better understanding

Backport to 10.9 and 10.8


mmattel (approved):

hodyroff:
Please use the wording "deprecated" only. Thats what we have in the release notes as well.
Deprecated means that we do not advise future usage and current usage must migrate ASAP.

mmattel:
> Please use the wording "deprecated" only. Thats what we have in the release notes as well. Deprecated means that we do not advise future usage and current usage must migrate ASAP.

Will file another PR to fix that.

PR #299, opened 2022-05-03
Adjust primary-replica words for redis

phil-davis:
Another change for issue https://github.com/owncloud/docs/issues/4721


mmattel:
@phil-davis we cant do that change here, needs to be done in core. Changes have been made recently in the config file in core that needs a CTD run anyways. I was just working on it but saw that some text stuff was not made well. I am going to fix this now in core and include that change too.

phil-davis:
good - I will close here and the change will come across later this day/week...

PR #357, opened 2022-05-17
shareType can't be optional if shareType is 1 or 0

jnweiger:
Language kniefielseling.
No backports needed.


mmattel:
So this is master only, no 10.10 and no 10.9 ?

mmattel:
After clarification BP to 10.10, 10.9 and 10.8

mmattel (approved):

PR #358, opened 2022-05-18
Fix command line display on SAML page

d7oc:
The additional bracket breaks the command line display


mmattel (approved):
Thanks a lot for catching this 👍
 
Backport to 10.10, 10.9and 10.8

PR #373, opened 2022-05-24
Temporary fix for the openSSL v3 incompatibility

mmattel:
This is a warning and description for a temporary fix for openSSL v3.

References:
https://github.com/owncloud/core/issues/40071 ([QA] Encryption cannot be enabled on Ubuntu 22.04)

Language review welcomed.

Note this PR complements, but is **not** dependent on the Pull Request U22.04 / PHP7.4  https://github.com/owncloud/docs-server/pull/369 (Server Preparation for Ubuntu 22.04) 

Backport to 10.10, 10.9 and 10.8

@pako81 @jnweiger @IljaN fyi

PR #376, opened 2022-05-24
[10.8] [PR 373] Temporary fix for the openSSL v3 incompatibility

mmattel:
Backport of #373


phil-davis (approved):
Seems relevant for 10.8 because people might could installing 10.8 on some later OS release, and come across this.

PR #386, opened 2022-06-01
More explicit description about platform config

d7oc:
Added a step for the platform configuration as this wasn't mentioned explicitly and I already faced wrong setups at this setup. It's important that is "Web" and nothing else for the browser to work.


mmattel:
Backport to 10.10 and 10.9

PR #394, opened 2022-06-02
Set a neutral ocis version

mmattel:
This sets the ocis version to `next` instead a real number. This is to satisfy the build process only and has no relevance here when things get changed in reality. The real version is set in the ocis and docs repo.

Antora hopefully gets once a includable attribute list or manifest eliminating this problem.

Backport to 10.10, 10.9 and 10.8


mmattel:
Will keep this as it is, closing

PR #428, opened 2022-06-15
Update using_letsencrypt.adoc

jnweiger:
to sudo or not to sudo...

I am not fully convinced, we should suggest snap to enterprise customers, but that is what upstream does these days .


mmattel:
Backport to 10.10 and 10.9

mmattel (approved):

EParzefall (approved):

PR #460, opened 2022-06-22
Added description of password-from-service-account

d7oc:
The description for the `--password-from-service-account` parameter was missing in the Password Options chapter


mmattel:
Backport to 10.10 and 10.9

PR #477, opened 2022-06-27
Updated section about encryption overhead

d7oc:
The encryption overhead changed over time as it was changed from BASE64 to binary in version 1.5.0. The updated documentation tries to reflect this.

Backport to 10.10 and 10.9


modules/admin_manual/pages/configuration/files/encryption/encryption_configuration.adoc
```diff
@@ -33,7 +33,8 @@ ownCloud’s server-side encryption encrypts files stored on the ownCloud server
 
 [NOTE]
 ====
-* Encrypting files increases their size by roughly 35%. Remember to take this into account when you are both provisioning storage and setting storage quotas.
+* Encrypting files increases their size by an 8KB header plus 100 bytes per each 8KB block of the file. Remember to take this into account when you are both provisioning storage and setting storage quotas.
+* If you are running an encryption module before version 1.5.0 or started with a version before 1.5.0 BASE64 encyption was used and might still be used for old files if untouched since then. In these cases old encrypted files increased their size by roughly 35%. We recommend to update the system.
```
mmattel, line 37:
```suggestion
* If you are running the encryption app which is delivered as part of the ownCloud system before version 1.5.0 or started with a version before 1.5.0, BASE64 encyption was used and might still be used for old files if untouched since then. In these cases, old encrypted files increased their size by roughly 35%. We recommend to update the ownCloud system to be able to use the benefit of reduced file sizes.
```

mmattel (approved):

PR #503, opened 2022-07-04
Improved sentence about PHP output buffering

d7oc:
Improved sentence about PHP output buffering as this is already the default in our tarball's `.htaccess`.


mmattel:
Backport to 10.10 and 10.9

mmattel:
Let's have an audio to discuss first

d7oc:
For fpm we also have a user.ini in the tarball. Or is there some other php variant next to apache and fpm?

DeepDiver1975:
Well. Apache without htaccess support and maintain all settings in Apache config itself.

Config space is endless 🤷

mmattel (approved):

PR #515, opened 2022-07-07
Make www-data a user in the redis group

jstell:
In order for owncloud to read the redis socket, it only needs to be added to the redis group. The redis user shouldn't be modified to make www-data its primary group.


phil-davis (approved):
This looks more correct to me.

@mmattel and anyone else, please comment.

mmattel (approved):

PR #530, opened 2022-07-14
Document that ownCloud 10.9 and above also works with Mariadb 10.7 and 10.8

phil-davis:
See core PRs https://github.com/owncloud/core/pull/40206 and https://github.com/owncloud/core/pull/40207 for a demonstration that Mariadb 10.7 and 10.8 work fine with ownCloud core. That should be true from ownCloud 10.9, when support for Mariadb 10.6 was explicitly added. There are no ownCloud core code changes needed for Mariadb 10.7 or 10.8.

IMO, we can mention that ownCloud core 10.9 and upwards supports using Mariadb all the way up to 10.8.

### DIFF
--- a/modules/admin_manual/pages/installation/system_requirements.adoc
+++ b/modules/admin_manual/pages/installation/system_requirements.adoc
@@ -38,7 +38,7 @@ For _best performance_, _stability_, _support_ and _full functionality_, we offi
 | {recommended-php-version}
 |===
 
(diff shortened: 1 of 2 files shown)


phil-davis:
Backport to 10.9 and 10.10 branches.

pmaier1 (approved):
Thanks for taking care :+1:

phil-davis:
Backports in PRs #531 and #532

PR #563, opened 2022-08-06
Update generated volume name in index.adoc

msteiger:
Having followed the instructions and using the provided `docker-compose.yml` file, the generated volume name is different from what is mentioned in the docs.

[modules/admin_manual/examples/installation/docker/docker-compose.yml](https://github.com/owncloud/docs-server/blob/master/modules/admin_manual/examples/installation/docker/docker-compose.yml)

My docker version is "Docker version 20.10.12, build 20.10.12-0ubuntu4" if that matters.

```bash
$ docker volume ls
DRIVER    VOLUME NAME
local     owncloud-setup_files
local     owncloud-setup_mysql
local     owncloud-setup_redis
```

### DIFF
--- a/modules/admin_manual/pages/installation/docker/index.adoc
+++ b/modules/admin_manual/pages/installation/docker/index.adoc
@@ -192,10 +192,7 @@ To inspect the volumes run:
 docker volume ls
 ----
 
-The volume name depends on the project name which builds the first part of the volume and the name of the volume in the docker file. 
-It can be configured using the variable `COMPOSE_PROJECT_NAME` which can also be defined in a `.env` file. 
-If not specified, the directory in which docker-compose is executed will be used as a name.
-The naming pattern of the volume is `<COMPOSE_PROJECT_NAME>_<VOLUME_NAME>`.
+The volume name depends on the project name which builds the first part of the volume and the name of the volume in the docker file. The naming pattern of the volume is `<COMPOSE_PROJECT_NAME>_<VOLUME_NAME>`. An environment variable for `COMPOSE_PROJECT_NAME` can be set and also be defined in a `.env` file. If not specified, the directory in which docker-compose is executed will be used as a name.
 
 To export the files of the project "owncloud-docker-server" as a tar archive run:
 [source,docker]
--- a/modules/admin_manual/pages/installation/docker/index.adoc
+++ b/modules/admin_manual/pages/installation/docker/index.adoc
@@ -189,7 +189,7 @@ It is the admin's responsibility to make the files persistent.
 To inspect the volumes run:
 [source,docker]
 ----
-docker volume ls
+docker volume ls | grep files
 ----
 
 The volume name depends on the project name which builds the first part of the volume and the name of the volume in the docker file. The naming pattern of the volume is `<COMPOSE_PROJECT_NAME>_<VOLUME_NAME>`. An environment variable for `COMPOSE_PROJECT_NAME` can be set and also be defined in a `.env` file. If not specified, the directory in which docker-compose is executed will be used as a name.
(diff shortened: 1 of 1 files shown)


d7oc (changes requested):
The volume name depends on the `COMPOSE_PROJECT_NAME` which builds the first part of the volume and the name of the volume in the docker file. So it will differ by design as long as we don't set the `COMPOSE_PROJECT_NAME` in an `.env` file as the `COMPOSE_PROJECT_NAME` is build from the directory in which docker-compose is executed when not specified. That's by the way also the reason why you run into troubles when moving a docker-compose.yml file to another directory and try to execute it there. See https://docs.docker.com/compose/reference/envvars/#compose_project_name.
So the pattern for the volume will be < COMPOSE_PROJECT_NAME>_<VOLUME_NAME>. So we shouldn't modify the documentation like described in this patch, but I would vote for adding an `.env` file or the corresponding `-p` parameter to the `docker-compose` calls, so we fix the name.

msteiger:
Thank you for the clarification. I now understand how it works. In order to also help others, I summarized your explanation and put it in a short paragraph. It would be enough to clarify the naming pattern in the docs as a first step, imho. What do you think?

modules/admin_manual/pages/installation/docker/index.adoc
```diff
@@ -189,13 +189,18 @@ It is the admin's responsibility to make the files persistent.
...
 ----
-docker volume ls | grep ownclouddockerserver
+docker volume ls
 ----
 
-To export the files as a tar archive run:
+The volume name depends on the project name which builds the first part of the volume and the name of the volume in the docker file. 
+It can be configured using the variable `COMPOSE_PROJECT_NAME` which can also be defined in a `.env` file. 
+If not specified, the directory in which docker-compose is executed will be used as a name.
+The naming pattern of the volume is `<COMPOSE_PROJECT_NAME>_<VOLUME_NAME>`.
```
mmattel, line 198:
```suggestion
The volume name depends on the project name which builds the first part of the volume and the name of the volume in the docker file. The naming pattern of the volume is `<COMPOSE_PROJECT_NAME>_<VOLUME_NAME>`. An environment variable for `COMPOSE_PROJECT_NAME` can be set and also be defined in a `.env` file. If not specified, the directory in which docker-compose is executed will be used as a name.
```

mmattel:
@d7oc I have made a review of the text.
Can you pls have a look if it is technically correct. Based on your ok I will proceed.

modules/admin_manual/pages/installation/docker/index.adoc
```diff
@@ -189,13 +189,15 @@ It is the admin's responsibility to make the files persistent.
...
 ----
 
-To export the files as a tar archive run:
+The volume name depends on the project name which builds the first part of the volume and the name of the volume in the docker file. The naming pattern of the volume is `<COMPOSE_PROJECT_NAME>_<VOLUME_NAME>`. An environment variable for `COMPOSE_PROJECT_NAME` can be set and also be defined in a `.env` file. If not specified, the directory in which docker-compose is executed will be used as a name.
+
+To export the files of the project "owncloud-docker-server" as a tar archive run:
 [source,docker]
 ----
-docker run -v ownclouddockerserver_files:/mnt \
+docker run -v owncloud-docker-server_files:/mnt \
```
d7oc, line 200: Should be changed into the following as the name needs to found on each installation and is not fixed:

```suggestion
docker run -v <YOUR_DOCKER_VOLUME>:/mnt \
```

modules/admin_manual/pages/installation/docker/index.adoc
```diff
@@ -189,13 +189,15 @@ It is the admin's responsibility to make the files persistent.
 To inspect the volumes run:
 [source,docker]
 ----
-docker volume ls | grep ownclouddockerserver
+docker volume ls
```
d7oc, line 192: We can leave a `grep` for `files` here as only the prefix will change, the name `files` is given in the compose file.

```suggestion
docker volume ls | grep files
```

d7oc (dismissed):
I'm fine with that if the two suggestions are included. For a later stage we could also think about adding a standard `.env` file to the documentation or the `-p` parameter.

mmattel (approved):

mmattel:
Backport to 10.10 and 10.9

PR #573, opened 2022-08-12
Public link share of a single file can now have edit access

### LINKED ISSUE #572: Public link share of a single file can now have edit access
## WHAT Needs to be Documented?
See core PR https://github.com/owncloud/core/pull/40264
"feat: allow editing of public link shared single files"

This will be released in core 10.11 - so no backporting of docs needed.

## WHERE Does This Need To Be Documented (Link)?
TBD - I will look

## (Optional) What Type Of Content Change Is This? 
- [x] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [ ] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

phil-davis:
Fixes #572 

Note: the ownCloud core behavior when creating public link shares is that if I ask for permissions 3 (read and update) on a public link share API share creation request, then I actually get just permission 1 (read). That has been that way for "a long time" (tm), and maybe there are existing clients that "accidentally" send permissions 3 when actually they only expect to get a public link share with read permission. (Probably requests asking for permission 3 on a public link should have returned some 4xxx code to indicate that the request is not valid, but it's too late for that now)

So that behavior needs to be preserved.

To create a public link share of a single file, the client has to request permissions 15 (read, update, create, delete) to actually get permissions 3 (read, update). I have noted that in the documentation - it is a bit of a weird thing that you ask for permissions 15 and get 3, but if you ask for permissions 3, you get 1.

### DIFF
--- a/modules/developer_manual/examples/core/scripts/curl/create-share.sh
+++ b/modules/developer_manual/examples/core/scripts/curl/create-share.sh
@@ -6,9 +6,13 @@
 base_uri={oc-examples-server-url}
 API_PATH=ocs/v1.php/apps/files_sharing/api/v1/shares
 
-# Create a public link share with read permissions, named "paris photo"
+# Create a public link share of a single file with read permissions, named "paris photo"
 curl --user {oc-examples-username}:{oc-examples-password} "$base_uri/$API_PATH" \
      --data 'path=/Photos/Paris.jpg&shareType=3&permissions=3&name=paris%20photo'
+
+# Create a public link share of a single file with read and write permissions, named "Notes"
+curl --user {oc-examples-username}:{oc-examples-password} "$base_uri/$API_PATH" \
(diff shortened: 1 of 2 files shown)


phil-davis:
@DeepDiver1975 please comment if I have written rubbish above.

mmattel:
Backports to 10.10 and 10.9 necessary?

phil-davis:
> Backports to 10.10 and 10.9 necessary?

No backports - this is a new feature for 10.11

mmattel:
Will merge latest on tuesday or if there is some ok inbetween.

mmattel (approved):

PR #586, opened 2022-08-24
Update config_sample_php_parameters.adoc

hodyroff:
Fixed description to fit with release notes and the actual function.


phil-davis:
This needs to be done in core, then there is a process to update the published docs here.

See https://github.com/owncloud/core/pull/40314 - please review that.

phil-davis:
Done in #589

PR #598, opened 2022-09-02
Update certificates.adoc

rezman42:
ucr seems not to be available


mmattel:
@d7oc can you double check?

d7oc:
First of all: Which UCS version are we talking about? I don't have one installation at hand but according to their documentation even UCS 5 still has ucr: https://docs.software-univention.de/manual/5.0/en/ip-config/radius.html?highlight=ucr#service-specific-password

d7oc:
ucr is a command in UCS. Don't know why you are talking about Ubuntu here.

rezman42:
thank you for clarifying, i got this wrong -> withdrawing/ closing the PR

Issue #252, opened 2022-04-13
Attempt to document installation of PHP 7.4 and necessary extensions on Ubuntu 22.04


pako81:
## WHAT Needs to be Documented?

Attempt to document installation of PHP 7.4 and necessary extensions on Ubuntu 22.04. 

1. Install SURY APT Repository: to install PHP 7.4 on Ubuntu 22.04 we will need to install SURY, third party repositories that provides PHP packages (we are going to use the Focal one as Jammy one is not available yet):

```
sudo add-apt-repository ppa:ondrej/php -y
sed -i 's/jammy/focal/' /etc/apt/sources.list.d/ondrej-ubuntu-php-jammy.list
sudo apt update
```

2. Running below command to install PHP7.4:

`sudo apt install php7.4`

returns an error:

```
The following packages have unmet dependencies:
 libapache2-mod-php7.4 : Depends: php7.4-cli but it is not going to be installed
                                             Depends: libssl1.1 (>= 1.1.0) but it is not installable
                                             Recommends: apache2 but it is not going to be installed
 php7.4-common : Depends: libssl1.1 (>= 1.1.0) but it is not installable
```

Problem —> In Ubuntu 22.04 `libssl1.1` has been superseded by `libssl3` and therefore following error is returned
Solution —> download `libssl1.1` package from http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2.12_amd64.deb and install it manually by running 

`sudo dpkg -i libssl1.1_1.1.1f-1ubuntu2.12_amd64.deb`

3. Install additional required modules:

```
sudo apt install php7.4-mysql, php7.4-mbstring, php7.4-redis, php7.4-imagick, php7.4-igbinary, php7.4-gmp, php7.4-bcmath, php7.4-curl, php7.4-gd, php7.4-zip, php7.4-imap, php7.4-ldap, php7.4-bz2, php7.4-ssh2, php-phpseclib, php7.4-common, php7.4-json, php7.4-xml, php7.4-dev, libsmbclient-dev, php-pear
```

4. Running below command to install the `php7.4-intl` package:

`apt install php7.4-intl`

returns an error:

```
The following packages have unmet dependencies:
 php7.4-intl : Depends: libicu66 (>= 66.1-1~) but it is not installable
```

Problem —> In Ubuntu 22.4 `libicu70` package is available
[...]

PR #599, opened 2022-09-02
Update ubuntu_22_04.adoc

rezman42:
php7.4-phpseclib seems not existing yet. using it here will break this easy-to-use package installation, leading to not installing any of these packages.

i proposed "php-phpseclib" since it was mentioned in the code block here:
https://github.com/owncloud/docs-server/issues/252#issuecomment-1116232718


mmattel (approved):
@rezman42 thanks for opening this PR. Your suggestion is correct. Merging.
Note that our [detailed installation guide](https://doc.owncloud.com/server/next/admin_manual/installation/manual_installation/server_prep_ubuntu_22.04.html) has the correct naming. The typo just sneaked into the quick guide.

PR #632, opened 2022-09-14
Update manual_upgrade.adoc

rezman42:
maintenance mode needs to be disabled to perform `occ upgrade`, if i didn't miss something:

```
:~# sudo -uwww-data ./occ ma:mo --on
Maintenance mode enabled
Please also consider to stop the web server on all ownCloud instances
:~# sudo -uwww-data ./occ upgrade
ownCloud is in maintenance mode - no app have been loaded

<warning>ownCloud is in maintenance mode</warning>
Maybe an upgrade is already in process. Please check the logfile (data/owncloud.log). If you want to re-run the upgrade procedure, remove the "maintenance mode" from config.php and call this script again.
:~# 
```

"Disable Maintenance Mode" can also be put before "Finalize the Upgrade"

### DIFF
--- a/modules/admin_manual/pages/maintenance/upgrading/manual_upgrade.adoc
+++ b/modules/admin_manual/pages/maintenance/upgrading/manual_upgrade.adoc
@@ -234,9 +234,18 @@ If you have configured a script for xref:installation/manual_installation/script
 
 == Finalize the Upgrade
 
+=== Disable Maintenance Mode
+
+Assuming your upgrade succeeded, disable maintenance mode using the occ command.
+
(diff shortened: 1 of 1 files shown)


mmattel:
sorry, this was closed accidentially, will take a look

rezman42:
thank you for re-opening and checking. i think i missed something. for the manual upgrade the article is right.

mmattel:
When upgrading an owncloud instance with the command `occ upgrade`, the instance must be in maintenance mode. At least this was the case since the last 8 years I know ownCloud. Pls recheck.

PR #658, opened 2022-09-28
doc update for owncloud/core#40389

mrow4a:
Adding some comments in `s3_compatible_object_storage_as_primary.adoc`

Backport to 10.11 and 10.10


mmattel (approved):

mrow4a:
@mmattel I am not sure, this feature is not yet available in oc10, it would be in next release, this is 10.12

mmattel:
Just for clarification, we should revert the backport to 10.11 and 10.10...?
We can do that easily, we just need to know 😄

mrow4a:
@mmattel I think yes, otherwise it could be confusing.

Issue #742, opened 2022-11-17
occ files:scan doesn't work on S3


voroyam:
## WHAT Needs to be Documented?
`occ files:scan` is not applicable on S3 Storages (files_primary_s3)

## WHERE Does This Need To Be Documented (Link)?
https://doc.owncloud.com/server/next/admin_manual/configuration/server/occ_command.html#the-filesscan-command

## WHY Should This Change Be Made?
for customers to know and understand that as of now we don't have a tool to sync up the s3 storage with the db

## (Optional) What Type Of Content Change Is This? 
- [x] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

PR #743, opened 2022-11-21
Update files:scan not to be used with S3

mmattel:
Fixes: #742 (occ files:scan doesn't work on S3)

Add a note to the occ files:scan command not to be used with S3.

Backport to 10.11 and 10.10

### DIFF
--- a/modules/admin_manual/pages/configuration/server/occ_commands/core_commands/_file_commands.adoc
+++ b/modules/admin_manual/pages/configuration/server/occ_commands/core_commands/_file_commands.adoc
@@ -131,7 +131,7 @@ The `files:scan` command
 
 File scans can be performed per-user, for a space-delimited list of users, for groups of users, and for all users.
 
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/configuration/server/occ_commands/core_commands/_file_commands.adoc
```diff
@@ -131,6 +131,8 @@ The `files:scan` command
 
 File scans can be performed per-user, for a space-delimited list of users, for groups of users, and for all users.
 
+NOTE: Scanning is only possible when using POSIX filesystems but not for objectstores like S3. This is because the objectstore implementation uses the database as primary datasource and the S3 storage for only holding the data blobs but no metadata. A sync from S3 storage to database is therefore not reasonable.
```
EParzefall, line 134:
```suggestion
NOTE: Scanning is only possible when using POSIX filesystems but not for object storages like S3. This is because the object storage implementation uses the database as primary data source and the S3 storage for only holding the data blobs but no metadata. A sync from S3 storage to database is therefore not reasonable.
```

EParzefall (approved):

voroyam:
Great work!

Issue #771, opened 2022-12-14
[QA] guests app blocklist description missing


jnweiger:
## WHAT Needs to be Documented?
https://doc.owncloud.com/server/next/admin_manual/configuration/user/guests_app.html#configuration
has a screenshot that shows a domain blocklist, but the text discussing this screenshot does not mention it.

The entries in the blocklist are exact matches (as per guests 0.12.2). That means, that an entry `example.com` blocks user@example.com, but user@mail.example.com is not blocked. admins are advised to list all possible subdomains explicitly.

Up to guests 0.12.1 the blocklist entries were a suffix match. So that an entry `example.com` would also block user@otherexample.com -- this was considered an error, and admins relying on this feature must please review their blocklists.

## WHERE Does This Need To Be Documented (Link)?
found the above link in the admin manual. maybe there is more...

## WHY Should This Change Be Made?
its incomplete. and the behaviour changed with te recent 0.12.2 release.
https://github.com/owncloud/guests/pull/529

## (Optional) What Type Of Content Change Is This? 
- [x] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [x] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [ ] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

PR #772, opened 2022-12-14
Update of the guest app description

mmattel:
Fixes: #771 ([QA] guests app blocklist description missing)

Adding proper descriptions when using blocklists

Backport to 10.11 and 10.10

### DIFF
--- a/modules/admin_manual/pages/configuration/user/guests_app.adoc
+++ b/modules/admin_manual/pages/configuration/user/guests_app.adoc
@@ -19,6 +19,12 @@ Install and enable the {oc-marketplace-url}/apps/guests[Guests] app if not alrea
 
 Check your Guests app's configuration in menu:Settings[Admin > Sharing]. There you can change the Guest's **group name** and add to or exclude apps from the app **whitelist** of the Guests app. Guests cannot access apps that are not on that list.
 
+With a blocklist, an admin can block domains for guest invitations:
+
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/configuration/user/guests_app.adoc
```diff
@@ -19,6 +19,12 @@ Install and enable the {oc-marketplace-url}/apps/guests[Guests] app if not alrea
 
 Check your Guests app's configuration in menu:Settings[Admin > Sharing]. There you can change the Guest's **group name** and add to or exclude apps from the app **whitelist** of the Guests app. Guests cannot access apps that are not on that list.
 
+With a blocklist, an admin can block domains for guest invitations:
+
+* Up to Guests 0.12.1, the blocklist entries were a suffix match. An entry like `example.com` would also block `user@otherexample.com` -- this was considered an error and admins relying on this feature must review their blocklists when upgrading to Guests 0.12.2.
+
+* Starting with Guests 0.12.2, the entries in the blocklist are exact matches. That means, that an entry `example.com` blocks `user@example.com`, but `user@mail.example.com` is *not* blocked. Admins are advised to list all possible subdomains explicitly.
```
phil-davis, line 26: and this behavior is being discussed in issue #541 - maybe the behavior will be changed again so that blocking "example.com" will block that whole domain (block example.com sub.example.com s1.s2.example.com ...). I think that that would be a more "logical" behavior.

But for now, what is written here is what happens.

phil-davis (approved):

modules/admin_manual/pages/configuration/user/guests_app.adoc
mmattel: True, when an update is coming we will add a new item and describe the changes.

PR #831, opened 2023-01-31
Update config_apps_sample_php_parameters.adoc

jnweiger:
removed cascading auto-provioning ...

According to dev docs https://github.com/owncloud/openidconnect#setup-auto-provisioning-mode

Backport to 10.11 and 10.10


mmattel (approved):

mmattel:
@DeepDiver1975 thanks a lot pointing to it. I will fix that in core and do a config-to-docs run afterwards

jnweiger:
Same issue with 
https://github.com/owncloud/core/blob/master/config/config.apps.sample.php#L390

(but I am not sure, if my proposed syntax is correct. Untested)

Issue #661, opened 2022-09-30
PHP 7.3 support will get dropped - needs documentation


mmattel:
## WHAT Needs to be Documented?
Removal of PHP7.3 support

## WHERE Does This Need To Be Documented (Link)?
https://doc.owncloud.com/server/next/admin_manual/installation/system_requirements.html
https://doc.owncloud.com/server/next/admin_manual/installation/manual_installation/manual_installation_prerequisites.html

## WHY Should This Change Be Made?
References: https://github.com/owncloud/core/pull/40394 ([full-ci] Drop PHP 7.3)

## (Optional) What Type Of Content Change Is This? 
- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [x] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

@phil-davis fyi

mmattel:
@phil-davis I "guess" that php7.3 support has been removed from core and will be published in 10.12, means we need to adapt the `antora.yml` file in this repo/master. Currently we have:

```
asciidoc:
  attributes:
    minimum-php-version: '7.3'
    minimum-php-printed: '7.3.0'
    minimum-php-version-short-code: '73'
    recommended-php-version: '7.4'
    recommended-php-version-short-code: '74'
    supported-php-versions: '7.3 and 7.4'
```
Can you advice?

phil-davis:
Yes, PHP 7.3 support is dropped in the upcoming core 10.12 release.

core 10.12 must run PHP 7.4

PR #835, opened 2023-01-31
Update PHP minimum version

mmattel:
Fixes: #661 (PHP 7.3 support will get dropped - needs documentation)

Update the minimum required PHP version to 7.4.

Locally tested, looks fine.

No backport !

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -7,9 +7,9 @@ nav:
 
 asciidoc:
   attributes:
-    minimum-php-version: '7.3'
-    minimum-php-printed: '7.3.0'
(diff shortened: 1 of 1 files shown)


EParzefall (approved):

phil-davis (approved):
Good to do this now, then when we make a 10.12 branch in the near future it will be like this.

mmattel:
I am preparing like hell for that.
As far I can see, no more open 10.12 doc-server issues - until one raises another one. We finally need to do the branching, but this will be done close to publish. Only the release notes in docs need to be updated.

PR #847, opened 2023-02-13
Also mention user guide?

jnweiger:
Should the entry page of the server guide only address admins and devs? I'd assume there are at least as many end users who end up here...

Backport to 10.10 and 10.11


mmattel (approved):

jnweiger (commented):
thanks!

PR #870, opened 2023-03-06
remove extra backslash

voroyam:
if you test it and copy paste the block - you get this: 

```
....
php7.4-redis php7.4-ldap php-phpseclib\
>
```
without the extra back slash the command gets executed as expected.


mmattel:
Thanks for catching this 👍

PR #912, opened 2023-03-30
Update encryption_configuration.adoc

jnweiger:
deprecate user-key-based encryptin harder.
Actually we should find all mentions of encryption:select-encryption-type user-keys or description of its web UI, it no longer exists in 10.12.1 / encryption 1.6.0 Upgrading from an earlier version, where user-key-based encryption was enabled, is currently harmless. It will continue to work.  But user-key-based encryption is deprecated, unmaintained and no longer tested. If it is still in use, migrate to master-key-ased encryption asap.

UPDATE by @mmattel 
Needs backport to 10.12


mmattel (approved):
LGTM

jnweiger:
Ouch I was to quick here. We won't ship this change with 10.12.1 -- planned for 10.13.0 instead.

PR #915, opened 2023-03-31
Revert 913

jnweiger:
Revert backport too.

I was too fast. the change was discussed and considered to risky for a patchlevel release.
This encryption-1.6.0 update will occur with 10.13.xx while 10.12.1 will keep encryption-1.5.3


mmattel:
Killing this one...

PR #919, opened 2023-04-04
Update ui-testing.adoc

jnweiger:
First we explain that chrome 74 is needed, and not 75 from latest, then we download exactly the right docker image, and finally we still run latest, despite all the nice preparations?? Naah!

 Wasted some hours there, today.

### DIFF
--- a/modules/developer_manual/pages/testing/ui-testing.adoc
+++ b/modules/developer_manual/pages/testing/ui-testing.adoc
@@ -62,7 +62,7 @@ taken to do the test.
 
 [source,console]
 ----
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
Correct - the PHP Behat/Mink/Selenium automated testing stack for the oC10 classic web UI only supports an older test interface to the browsers. It still does not support the newer browser test protocols. See issue https://github.com/owncloud/core/issues/37113

We will look again to see if there is a way forward with this, but I haven't noticed anything useful change in Behat/Mink/Selenium recently.

PR #934, opened 2023-04-11
added instructions how to enable legacy support

voroyam:
Added an example to to implement the fix required for current LTS version of Ubuntu (22.04) and ownCloud encryption.

relevant docs page: 
https://doc.owncloud.com/server/next/admin_manual/installation/manual_installation/manual_installation_prerequisites.html#openssl-version

it's in the "important" block.

please add the docs magic to it so it looks nice. The commands should be in code blocks, but I did not know how to do them right.

I am aware that this example might become outdated in the future, but without it many users and or admins might get lost on what to implement where and how. This should act as a example. 

I also added an anchor for the relevant section of the ssl docs.


mmattel:
Fixed some layouting
Fixed why ever the anchor in the attribute did not work so I added the link directly

Backport to 10.11 and 10.12

PR #945, opened 2023-04-17
Update theming.adoc

voroyam:
removed occ prefix. Quick guide expects one to be in the apps-external directory, the occ command + prefix does not work. Additionally the occ wrapper is expected at this point.


mmattel:
Disagreed, old story, the quickguide is a special case, all occ commands are always complete, closing.

voroyam:
then the instructions in the theming guide need to be changed. you can't have cd /apps-external and then on the next line have .occ because there is no occ script in /apps-external so you run in to an error.

Additionally - occ wrapper has been a part of the containers for years so the quick guide is not a special case.

PR #951, opened 2023-04-19
Update collabora_secure_view.adoc [10.12.0]

pako81:
Preview role for public links does not exist yet. Will likely be implemented in 10.13.0. References to it should be removed from the 10.12 documentation. "Next" can stay.


mmattel (approved):
Approving.

PR #972, opened 2023-04-28
Added Redis requirements

d7oc:
This PR adds the system requirements Redis

### DIFF
--- a/modules/admin_manual/pages/installation/system_requirements.adoc
+++ b/modules/admin_manual/pages/installation/system_requirements.adoc
@@ -31,6 +31,9 @@ For _best performance_, _stability_, _support_ and _full functionality_, we offi
 | Database
 | xref:#database-requirements[MariaDB] 10.5 ^1^
 
+| Redis
+| >= 6
+
 | Web server
 | Apache 2.4 with xref:installation/manual_installation/manual_installation.adoc#configure-the-web-server[`prefork and mod_php`]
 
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/installation/system_requirements.adoc
```diff
@@ -31,6 +31,9 @@ For _best performance_, _stability_, _support_ and _full functionality_, we offi
 | Database
 | xref:#database-requirements[MariaDB] 10.5 ^1^
 
+| Redis
+| >= 6
+
```
pako81, line 36: We should probably specify this is only in case connection parameters and SSL/TLS support is needed.

d7oc: If the specify this the audience might raise the question "what is the requirements if the support is not needed?". So what would be our answer to this. We should add this as well upfront.

mmattel:
* What is the status, ist all resolved or are more changes needed.
* Backports to 10.12 and 10.11 ?

modules/admin_manual/pages/installation/system_requirements.adoc
pako81: I don't think there are any specific requirements related to the redis server version in case support for connection parameters in combination with SSL/TLS is not required.

d7oc: Then let's simplify it to >=6 in general, even if it only applies to a specific condition. Older versions of Redis aren't supported with security fixes anyway. See https://redis.io/docs/about/releases/.

d7oc:
@mmattel Don't know if we have some "documentation calendar" but we should but a note somewhere, that whenever Redis releases a new major version this part of the documentation needs an overhaul (once added).

mmattel (approved):

PR #980, opened 2023-05-15
Update metrics.adoc for PR #166

jnweiger:
Fixes: https://github.com/owncloud/metrics/issues/170 ([QA] Data format of metrics/api/v1/metrics changed)

We now mention new the fields usedQuota and usedOther --  maybe that is a bit to much, as all other fields are not explicitly mentioned with their exact names...

If you ask me, it does not harm :-)

### DIFF
--- a/modules/admin_manual/pages/enterprise/reporting/metrics.adoc
+++ b/modules/admin_manual/pages/enterprise/reporting/metrics.adoc
@@ -46,7 +46,7 @@ The following data is available for analysis:
 .**System data**
 * Date/Time stamp - Server time of the request
 * Storage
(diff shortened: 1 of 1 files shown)


jnweiger:
This depends on the having metrics app 1.3.0 -- which is not bundled with 10.12 or earlier, but compatible. 
not sure if this qualifies for backports or not.

mmattel (approved):

mmattel:
> not sure if this qualifies for backports or not.

It does, because we always describe the latest app version. Backporting to the version this app is supported =
Backporting to 10.12 and 10.11

PR #998, opened 2023-06-07
improve wording of some Scenario examples

koebel:


phil-davis (approved):
well spotted, thank you.

This is developer/QA documentation. So it is fine to stay just in master. There is no need to backport it to 10.12 or 10.11

PR #600, opened 2022-09-05
Add recent core changes

mmattel:
Recent changes from core.

No Backport

@pmaier1

PR #1005, opened 2023-06-15
Fix missing image (10.11 only)

mmattel:
This is a 10.11 fix ony.

For what ever reason, this image was missing and another one was not used. Maybe an issue from earlier times.


phil-davis (commented):
The core release notes say that 
Enhancement - Persistent major file version workflow: [#40531](https://github.com/owncloud/core/pull/40531)
was released in 10.12.0

So this image is not correct for 10.11

Maybe something was accidentally backported to 10;11 that caused this image to be referenced? And so that accidental backport should be reverted.

mmattel:
> Maybe something was accidentally backported

I did not find a reference to this image in any page of 10.11
With the upcoming 10.13 this release is no longer supported or visible in docs. Is it really worth to search for the cause?

mmattel:
I did some investigation:
`git log --follow --diff-filter=A -- **concious-persistent-major-versions.png`
--> `commit 7d67bd2981ad162fa80621a7783fa45afc1af2d2`

Then searching for this commit in GH, I get PR #600 - which is master only and has no backports...
Furthermore, checking page content, the differences between master and 10.11 are correct.

I am merging therefore.

EParzefall (approved):

PR #1021, opened 2023-07-06
Update oidc.adoc

nicholas-wilson-au:
I tried the original  \ and it did not work.  I had to change it to / to get working.

Backport to 10.11 and 10.12


mmattel (approved):
@nicholas-wilson-au thanks for catching 👍 
`\.` escapes the dot but the rewrite rule must apply to `/.well-known/...`

PR #1029, opened 2023-07-10
Update config_sample_php_parameters.adoc

pako81:
Update description of `share_folder` config.php parameter for guest users. Related to https://github.com/owncloud/core/pull/40864


mmattel (changes requested):
You **MUST** do the change in your core PR adapting the config.sample file.

I will then, when merged, do the `config-to-docs` run to transport the changes to this repo.
If there are any changes added to a config.sample here, they will be eliminated on the next `config-to-docs` run.

pako81:
Ah right, thanks for clarifying it. I have done the changes to config.sample.php directly in the core PR.

PR #1036, opened 2023-07-12
warn about external storages in guests_app.adoc

jnweiger:
Guests have access to external storages, even when the files_external app is not in the whitelist.
I could not find anything to limit external storages to non-guest users. IMHO, that would be a sane default.

@pako81 Maybe this is a bug and should be fixed, rather than documented?

Not sure, if this warning should be here, or in the section about external storages instead.

### DIFF
--- a/modules/admin_manual/pages/configuration/user/guests_app.adoc
+++ b/modules/admin_manual/pages/configuration/user/guests_app.adoc
@@ -19,6 +19,8 @@ Install and enable the {oc-marketplace-url}/apps/guests[Guests] app if not alrea
 
 Check your Guests app's configuration in menu:Settings[Admin > Sharing]. There you can change the Guest's **group name** and add to or exclude apps from the app **whitelist** of the Guests app. Guests cannot access apps that are not on that list.
 
+Note: Per default guests see external storages that are available to all users (i.e. empty user/group list)
+
 With a blocklist, an admin can block domains for guest invitations:
 
 * Up to Guests 0.12.1, the blocklist entries were a suffix match. An entry like `example.com` would also block `user@otherexample.com` -- this was considered an error and admins relying on this feature must review their blocklists when upgrading to Guests 0.12.2.


pako81:
This is happening because `files_external`, `files_external_dropbox`, `files_external_ftp`, `files_external_s3` and `windows_network_drive` are included in the `CORE_WHITELIST` https://github.com/owncloud/guests/blob/master/lib/AppWhitelist.php#L33 and are therefore hard-coded.

I remember we discussed this some time ago as we updated the whitelist for the guests app for version 0.12.1. And, if I recall correctly, the decision at that time was to leave external storages apps.

I am also not a big fan of letting guests access external storages per default. Note that if we decide to change this, the guests will still be able to see the external storage listed but as soon as they click on it an error will be returned:

image

I am open for changing current behaviour. @phil-davis @jvillafanez what do you think?

phil-davis:
Someone needs to decide what the requirement is, and then sort out the effort required if a change to the current behavior is needed.

For example, if a company has some external storage that is available to "all users" then probably the storage contains company-wide documents - maybe policies/procedures/forms etc. And I expect that they would not want or expect that guest users can view such things. So they would want a way to control that. Maybe they can already? Mount the external storage to a group that contains "all staff"?

mmattel:
Should this PR now be continued (review/merge/beackport) or is it on-hold? For the latter, put it to draft.

pako81:
Discussed with @hodyroff. Guests should not access external storages normally available for "all" users.

We apparently have two options here:

1. still show external storages to guests but as soon as they click on them they will get an "storage is temporary not available" error (it may however be confusing for the guests)

2. not display external storages at all to guests, more complicated as front-end code would be required.

Probably option 1. would already suffice here.

mmattel:
This is going to be a technical discussion that should be held in core.
Whatever the finding is to be documented, the current changes or new stuff, pls drop a doc related comment. Not working on it until then.

jnweiger:
Noting todo for documentation here right now. Needs to be sorted out in a core or enterprise ticket first.

jnweiger:
Followup in https://github.com/owncloud/core/issues/40894

PR #1037, opened 2023-07-13
Update impersonate_users.adoc

jnweiger:
Never logged in -> no impersonate!

But are we sure, that impersonate is incompatible with shibboleth or openidconnect? I never noticed that...

### DIFF
--- a/modules/admin_manual/pages/configuration/general_topics/impersonate_users.adoc
+++ b/modules/admin_manual/pages/configuration/general_topics/impersonate_users.adoc
@@ -12,6 +12,7 @@
 
 * This functionality is available to administrators only.
 * Impersonating is not possible when using Shibboleth or OpenID.
+* Impersonating is not possible when the specific user has never logged in before.
 
 == Impersonating a User
 


cdamken:
@jnweiger I don't want to be a smart-ass, but IMHO the word `before` is not needed 
```
Impersonating is not possible when the specific user has never logged in.
```

jnweiger:
That is fine with me either way. My idea was, that redundant language could server as an emphasis.

A language reviewer might also wonder, why we have a gerund in "Impersonating is not possible ..."
"Impersonate is not possible ..." should work the same. IMO.

cdamken:
> But are we sure, that impersonate is incompatible with shibboleth or openidconnect? I never noticed that...

how about this part?

jnweiger:
> > But are we sure, that impersonate is incompatible with shibboleth or openidconnect? I never noticed that...
> 
> how about this part?

At least openid works fine, IMO -> https://github.com/owncloud/docs-server/pull/1038

mmattel (approved):

PR #1038, opened 2023-07-13
Update impersonate_users.adoc opendconnect works

jnweiger:
impersonate definitly works with openidconnect.
I cannot imagine a reason, why it should not work. Tested. Works fine.

It probably even works with shibleth. But I cannt easly test that.

### DIFF
--- a/modules/admin_manual/pages/configuration/general_topics/impersonate_users.adoc
+++ b/modules/admin_manual/pages/configuration/general_topics/impersonate_users.adoc
@@ -11,7 +11,7 @@
 == Limitations
 
 * This functionality is available to administrators only.
-* Impersonating is not possible when using Shibboleth.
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/configuration/general_topics/impersonate_users.adoc
```diff
@@ -11,7 +11,7 @@
 == Limitations
 
 * This functionality is available to administrators only.
-* Impersonating is not possible when using Shibboleth or OpenID.
+* Impersonating is not possible when using Shibboleth.
```
mmattel, line 14:
```suggestion
* Impersonating may not work when using Shibboleth.
```
Relaxing the statement according the comment of @jnweiger - ok?

mmattel (approved):

Issue #1044, opened 2023-07-24
Upgrade - Docker - Mariadb


voroyam:
## WHAT Needs to be Documented?
`MARIADB_AUTO_UPGRADE=1 ` has to be configured when upgrading ownClouds for the DB to be upgraded accordingly.

## WHERE Does This Need To Be Documented (Link)?
https://doc.owncloud.com/server/next/admin_manual/maintenance/upgrading/manual_upgrade.html#troubleshooting

## WHY Should This Change Be Made?
because if you don't you get these kind of errors after the upgrade:

(example upgrade from 10.6 to 10.11)

```
owncloud_mariadb  | 2023-07-24  8:14:10 4 [ERROR] Incorrect definition of table mysql.column_stats: expected column 'histogram' at position 10 to have type longblob, found type varbinary(255).
owncloud_mariadb  | 2023-07-24  8:14:10 4 [ERROR] Incorrect definition of table mysql.column_stats: expected column 'hist_type' at position 9 to have type enum('SINGLE_PREC_HB','DOUBLE_PREC_HB','JSON_HB'), found type enum('SINGLE_PREC_HB','DOUBLE_PREC_HB').
```

CC @d7oc

mmattel:
Where (not in docs) needs `MARIADB_AUTO_UPGRADE=1` to be set and what happens if the upgrade was not made with that key? Can it be post- added and the error "goes away"?

d7oc:
Should be added to the docker compose example in https://doc.owncloud.com/server/next/admin_manual/installation/docker/#docker-compose-yaml-file

It can also be added afterwards to make the error go away.

mmattel:
> Should be added to the docker compose example

And if one is not using docker?

voroyam:
> > Should be added to the docker compose example
> 
> And if one is not using docker?

as stated in the title of the issue, this only happens when using docker

d7oc:
> And if one is not using docker?

This is a docker ENV variable don't know how it will work on bare metal.

PR #1045, opened 2023-07-24
Added MARIADB_AUTO_UPGRADE=1

d7oc:
Fixes: #1044

Backport to 10.11 and 10.12

### DIFF
--- a/modules/admin_manual/examples/installation/docker/docker-compose.yml
+++ b/modules/admin_manual/examples/installation/docker/docker-compose.yml
@@ -48,6 +48,7 @@ services:
       - MYSQL_USER=owncloud
       - MYSQL_PASSWORD=owncloud
       - MYSQL_DATABASE=owncloud
+      - MARIADB_AUTO_UPGRADE=1
     command: ["--max-allowed-packet=128M", "--innodb-log-file-size=64M"]
     healthcheck:
       test: ["CMD", "mysqladmin", "ping", "-u", "root", "--password=owncloud"]
--- a/modules/admin_manual/pages/maintenance/upgrading/database_upgrade.adoc
+++ b/modules/admin_manual/pages/maintenance/upgrading/database_upgrade.adoc
@@ -9,7 +9,7 @@ Usually, when upgrading a database, follow the guides the vendor provides. You m
 
 This document supports an upgrade in case there are prerequisites and/or special steps to be taken to upgrade to a particular database version.
 
-== Upgrading to MariaDB 10.6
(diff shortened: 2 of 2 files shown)


mmattel:
According to: `MariaDB/mariadb-docker/issues/350`, you need to run
`mysql_upgrade` or `mariadb-upgrade` as part of the upgrade procedure when using a bare metal environment. `MARIADB_AUTO_UPGRADE` is the docker version of it and for MariaDB only.

To have a more neutral approach, see: [Upgrading Databases in Docker](https://theorangeone.net/posts/upgrading-docker-databases/). This would be inline for both bare-metal and docker as it uses the same underlaying command.

d7oc:
IMHO this is contradictory to the container approach if you start executing arbitrary commands in containers. Thus the correct way would be the ENV setting for docker.

mmattel:
> Thus the correct way would be the ENV setting for docker.

Fully agreed. What is the envvar for mysql (if one exists, I have not found one)?
We can use the envvar for MariaDB and use the docker command showing an alternative for mysql. AND adding the plain command to the bare-metal environment.
No-go: MariaDB only, no bare-metal.
Comments and ideas welcomed.

d7oc:
> Fully agreed. What is the envvar for mysql (if one exists, I have not found one)?

Seems for MySQL this isn't required according to this article: https://dev.mysql.com/doc/refman/8.0/en/docker-mysql-getting-started.html#docker-upgrading

So it will just automatically update the database without having a special ENV var set.

mmattel:
> Seems for MySQL this isn't required according to this article:

reading the note in the article just a bit below:
...
For MySQL 8.0.15 and earlier: You need to complete the upgrade process by running the [mysql_upgrade](https://dev.mysql.com/doc/refman/8.0/en/mysql-upgrade.html) utility in the MySQL 8.0 Server container (the step is not required for MySQL 8.0.16 and later):
...

d7oc:
I missed that, but given the support status: https://endoflife.date/mysql I would add just a disclaimer that users should use 8.0.16 or later instead of describing this.

Side note: If you set the 8.0 tag in MySQL you will get 8.0.34 right now anyway: https://hub.docker.com/layers/library/mysql/8.0/images/sha256-1da5cf22e7177925fa4d2a707936ca4ade59ba10b40b23550b0ffc12f8340600?context=explore

So I'm not sure if that needs to be written at all. As you explicitly need to pin an old version to run into this.

mmattel (approved):

PR #1078, opened 2023-08-16
Update federated_cloud_sharing_configuration.adoc

jnweiger:
trusted server setup gets more complex.

Hope I correclty misunderstood what Piotr found in https://github.com/owncloud/enterprise/issues/5676#issuecomment-1512708126

It must be very frustrating for a suer to distringuish between a trusted server entry and a remote trusted domains entry and then force run selected cron jobs manually...

(At least, I believe, I could obsolete direct database hacking here.)

Backport to 10.12 and master

PR #1087, opened 2023-08-17
Update federated_cloud_sharing_configuration.adoc

pako81:
Rework of https://github.com/owncloud/docs-server/pull/1078

Backport to master only

### DIFF
--- a/modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
+++ b/modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
@@ -20,29 +20,26 @@ However, to do so, you have to set `'sharing.federation.allowHttpFallback' => tr
 
 Follow these steps to establish a trusted connection between two servers.
 
-. Verify that both servers have SSL certificates.
-  If you open the server URL in your browser and see a lock icon on the left-hand side of the address bar, the certificate is valid.
+. Verify that both servers have SSL certificates
+  If you open the server URL in your browser and see a lock icon on the left-hand side of the address bar, the certificate is valid
 +
 .Lock icon in the address bars in Firefox, Google Chrome, and Safari.
 image:configuration/files/browser-address-bars.png[Lock icon in Firefox, Google Chrome, and Safari]
-. Verify that the `'overwrite.cli.url' => 'https://<SERVER_URL>' setting is configured to the correct URL, instead of `localhost`, in **config.php**.
-. Reset the federation job in your `oc_jobs` table. 
-  This job is required to get the verification token from the other server to establish a federation connection between two servers. 
-  The resetting ensures that it will be executed when we run xref:configuration/server/background_jobs_configuration.adoc#cron[system:cron] later.
-+
-[source,bash]
-----
-mysql -u root -e "update oc_jobs set last_run=0 where class='OCA\\Federation\\SyncJob';" owncloud;
-mysql -u root -e "update oc_jobs set last_checked=0 where class='OCA\\Federation\\SyncJob';" owncloud;
-----
-. Navigate to **admin settings -> sharing -> Federation**
-. Add **server 1** to the trusted servers on **server 2**.
-. Add **server 2** to the trusted servers on **server 1**.
-. Now run the cron job in your ownCloud directory (for example `/var/www/owncloud/`).
+. Verify that the `'overwrite.cli.url' => 'https://<SERVER_URL>'` setting is configured to the correct URL instead of `localhost` in **config.php**
+. Navigate to **admin settings -> Sharing -> Federation**
+. Add **server 1** to the trusted servers on **server 2**
+. Add **server 2** to the trusted servers on **server 1**
+. Run `system:cron` in the ownCloud directory of both servers (for example `/var/www/owncloud/`)
 +
 [source,bash,subs="attributes+"]
 ----
 {occ-command-example-prefix} system:cron
+----
+. Run the `OCA\Federation\SyncJob` job with the `force` option (you can get the correspondent job ID by using the `background:queue:status` occ command)
++
+[source,bash,subs="attributes+"]
+----
+{occ-command-example-prefix} background:queue:execute --force 10
 ----
 . Now the check should be green
 . Sync now your users with
@@ -52,7 +49,7 @@ mysql -u root -e "update oc_jobs set last_checked=0 where class='OCA\\Federation
 {occ-command-example-prefix} dav:sync-system-addressbook
 {occ-command-example-prefix} federation:sync-addressbook
 ----
-. Configure automatic acceptance of new federated shares.
+. Configure automatic acceptance of new federated shares
 +
 NOTE: Automatic acceptance of new federated shares will not work if the option `Add server automatically once a federated share was created successfully` is also set. This is done because of security concerns.
 +
--- a/modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
+++ b/modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
@@ -46,7 +46,7 @@ You can get the corresponding job ID by using the `background:queue:status` occ 
 
 . The check should now be green
 
-. Sync now your users by running on both servers
+. Now sync your users by running on both servers
 +
 [source,bash,subs="attributes+"]
 ----
@@ -70,7 +70,7 @@ NOTE: Automatic acceptance of new federated shares will not work if the option `
 
 There are ownCloud instances that are not connected to the internet. They have no possibility to reach the public network. Therefore Federation will not work without a proxy.
 
-To set the `proxy` and `proxyuserpwd` configuration variables, in `config/config.php`. `proxy` sets the proxy’s hostname, and `proxyuserpwd` sets the username and password credentials, in `username:password` format.
+Set the `proxy` and `proxyuserpwd` configuration variables, in `config/config.php`. `proxy` sets the proxy’s hostname, and `proxyuserpwd` sets the username and password credentials, in `username:password` format.
 
 == Creating a New Federation Share
 
@@ -150,7 +150,7 @@ You can also configure these settings of the cronjob:
 
 === Persistent Locks Are Not Guaranteed
 
-There is a known bug propagated persistent locks to federated instances. If a user creates an exclusive lock on a share, no other users should be able to modify it, nor its contents, and all users should see a lock icon on the share.
+There is a known bug in propagation of persistent locks to federated instances. If a user creates an exclusive lock on a share, no other users should be able to modify it, nor its contents, and all users should see a lock icon on the share.
 
 However, this isn't the case. The following functionality has been recorded:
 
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
```diff
@@ -20,39 +20,36 @@ However, to do so, you have to set `'sharing.federation.allowHttpFallback' => tr
...
-. Navigate to **admin settings -> sharing -> Federation**
-. Add **server 1** to the trusted servers on **server 2**.
-. Add **server 2** to the trusted servers on **server 1**.
-. Now run the cron job in your ownCloud directory (for example `/var/www/owncloud/`).
+. Run the `OCA\Federation\SyncJob` job with the `force` option on both servers (you can get the correspondent job ID by using the `background:queue:status` occ command)
 +
 [source,bash,subs="attributes+"]
 ----
-{occ-command-example-prefix} system:cron
+{occ-command-example-prefix} background:queue:execute --force 10
```
jnweiger, line 42: Is that sufficient?
I am not sure, if maybe we still need to call occ system:cron 
t really execute the job?

pako81: Based on my tests it seems to be sufficient.

mrow4a: we cannot be sure that id 10 will be what we need. it varies per system

pako81: correct but there is also documented `you can get the correspondent job ID by using the background:queue:status occ command`.

jnweiger: How about something like:
```
fedJobId=$(occ background:queue:status | grep Federation.SyncJob | cut -d '|' -f 2)
occ background:queue:execute --force $fedJobId
```

jnweiger: I've checked a few of my systems, it is always 10 there. I really thought that is a constant. :frowning:

modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
```diff
@@ -20,39 +20,36 @@ However, to do so, you have to set `'sharing.federation.allowHttpFallback' => tr
...
 ----
-mysql -u root -e "update oc_jobs set last_run=0 where class='OCA\\Federation\\SyncJob';" owncloud;
-mysql -u root -e "update oc_jobs set last_checked=0 where class='OCA\\Federation\\SyncJob';" owncloud;
+{occ-command-example-prefix} system:cron
 ----
-. Navigate to **admin settings -> sharing -> Federation**
-. Add **server 1** to the trusted servers on **server 2**.
-. Add **server 2** to the trusted servers on **server 1**.
-. Now run the cron job in your ownCloud directory (for example `/var/www/owncloud/`).
+. Run the `OCA\Federation\SyncJob` job with the `force` option on both servers (you can get the correspondent job ID by using the `background:queue:status` occ command)
```
phil-davis, line 38:
```suggestion
. Run the `OCA\Federation\SyncJob` job with the `force` option on both servers (you can get the corresponding job ID by using the `background:queue:status` occ command)
```

modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
phil-davis, line 42:
```suggestion
{occ-command-example-prefix} background:queue:execute --force <jobId>
```

Can we just mention `<jobId>` something like this?
How do we usually indicate to users that they need to substitute their own value in a command?

jnweiger: It never works for me on the first attempt.
I have now succeeded in one case, where I did exactly as described here, 
got only a yellow marker in the webUI, then pondered a while, and re-ran (a few minutes later) 
`occ background:queue:execute --force $fedJobId`
This additional (or delayed) command then did the magic: It added the entry to oc_addressbooks, and the trusted server became green in the UI.

jnweiger: Also, I'd suggest to not only use `--force` here, (which does not force much, it still scares the user with a fat where he has to type 'yes') but use `--force --accept-warning` which really runs the command without asking questions.

pako81: `--force` is rather meant for running the job even if it is already within the time interval. So the time interval will be ignored and `last_run` will be set to `0`. But yes, also using `--accept-warning` makes sense.

mmattel: Mind to adapt @phil-davis suggestable commit with whatever option is correct or commit the suggestion if it is ok? Else this stays open and nobody cares...

modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
```diff
@@ -5,109 +5,109 @@
...
 ----
-{occ-command-example-prefix} system:cron
+{occ-command-example-prefix} background:queue:execute --force <jobId>
 ----
-. Now the check should be green
-. Sync now your users with
+
+. The check should now be green
+
+. Sync now your users by running on both servers
```
phil-davis, line 49:
```suggestion
. Now sync your users by running on both servers
```

Or leave out "now".

modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
```diff
@@ -5,109 +5,109 @@
...
 == Working With Proxies
 
-There are ownCloud instances that are not connected to the internet. 
-They have no possibility to reach the public network. 
-Therefore Federation will not work without a proxy. 
+There are ownCloud instances that are not connected to the internet. They have no possibility to reach the public network. Therefore Federation will not work without a proxy.
 
-To set the `proxy` and `proxyuserpwd` configuration variables, in `config/config.php`.
-`proxy` sets the proxy’s hostname, and `proxyuserpwd` sets the username and password credentials, in `username:password` format.
+To set the `proxy` and `proxyuserpwd` configuration variables, in `config/config.php`. `proxy` sets the proxy’s hostname, and `proxyuserpwd` sets the username and password credentials, in `username:password` format.
```
phil-davis, line 73:
```suggestion
Set the `proxy` and `proxyuserpwd` configuration variables, in `config/config.php`. `proxy` sets the proxy’s hostname, and `proxyuserpwd` sets the username and password credentials, in `username:password` format.
```

modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
```diff
@@ -150,11 +150,9 @@ Use the following command to force a run of the scanner cronjob:
 
 === Persistent Locks Are Not Guaranteed
 
-There is a known bug propagated persistent locks to federated instances.
-If a user creates an exclusive lock on a share, no other users should be able to modify it, nor its contents, and all users should see a lock icon on the share.
+There is a known bug propagated persistent locks to federated instances. If a user creates an exclusive lock on a share, no other users should be able to modify it, nor its contents, and all users should see a lock icon on the share.
```
phil-davis, line 153:
```suggestion
There is a known bug in propagation of persistent locks to federated instances. If a user creates an exclusive lock on a share, no other users should be able to modify it, nor its contents, and all users should see a lock icon on the share.
```

modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
```diff
@@ -5,109 +5,109 @@
...
-However, to do so, you have to set `'sharing.federation.allowHttpFallback' => true,` in **config/config.php**.
+IMPORTANT: We strongly recommend using HTTP for development and testing purposes. However, to do so, you have to set `'sharing.federation.allowHttpFallback' => true,` in **config/config.php**.
 
 == Configuration
 
 Follow these steps to establish a trusted connection between two servers.
 
-. Verify that both servers have SSL certificates.
-  If you open the server URL in your browser and see a lock icon on the left-hand side of the address bar, the certificate is valid.
+. Verify that both servers have SSL certificates. +
```
phil-davis, line 18: Is the "+" supposed to be there?

mmattel: Yes. It tells the (numbered ) list item that the current sentence ends here, but the next line(s) are part of the item and therefore properly intented. (soft break). I can make you a build on staging so you can "optical review" this. Before commiting this change, a final local build rendered fine.

mmattel (approved):

phil-davis (approved):

mmattel:
Changed a text based on a chat with @jnweiger

modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
```diff
@@ -20,39 +20,36 @@ However, to do so, you have to set `'sharing.federation.allowHttpFallback' => tr
...
 image:configuration/files/browser-address-bars.png[Lock icon in Firefox, Google Chrome, and Safari]
-. Verify that the `'overwrite.cli.url' => 'https://<SERVER_URL>' setting is configured to the correct URL, instead of `localhost`, in **config.php**.
-. Reset the federation job in your `oc_jobs` table. 
-  This job is required to get the verification token from the other server to establish a federation connection between two servers. 
-  The resetting ensures that it will be executed when we run xref:configuration/server/background_jobs_configuration.adoc#cron[system:cron] later.
+. Verify that the `'overwrite.cli.url' => 'https://<SERVER_URL>'` setting is configured to the correct URL instead of `localhost` in **config.php**
+. Navigate to **admin settings -> Sharing -> Federation**
+. Add **server 1** to the trusted servers on **server 2**
+. Add **server 2** to the trusted servers on **server 1**
+. Run `system:cron` from the ownCloud directory of both servers (for example `/var/www/owncloud/`)
```
jnweiger:
```suggestion
. Run `{occ-command-example-prefix} system:cron` in both servers
```

Wrong place to discuss if it is needed to really chdir somwhere, when running occ commands. (And it should not be needed to chdir in anycase, imo, I never do)

But occ should be mentioned, when running occ commands.

jnweiger (changes requested):

modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
mmattel: I have further improved the text based on your changes, see commit, pls approve

jnweiger (approved):
LGTM

PR #1154, opened 2023-10-27
post_max_size reduced from 16G to 10M

EParzefall:
Reducing post_max_size to a more reasonable value.
Backport to 10.12 and 10.13 necessary


modules/admin_manual/pages/configuration/files/big_file_upload_configuration.adoc
```diff
@@ -64,7 +64,7 @@ Set the following parameters inside the corresponding file using your own desire
 {userini}::
 [source,php]
 ----
-post_max_size=16G
+post_max_size=10G
```
EParzefall, line 67:
```suggestion
post_max_size=10M
```

modules/admin_manual/pages/configuration/files/big_file_upload_configuration.adoc
```diff
@@ -73,7 +73,7 @@ upload_tmp_dir=/mnt/php_big_temp/
 {htaccess}::
 [source,php]
 ----
-php_value	post_max_size=16G
+php_value	post_max_size=10G
```
EParzefall, line 76:
```suggestion
php_value	post_max_size=10M
```

modules/admin_manual/pages/configuration/files/big_file_upload_configuration.adoc
```diff
@@ -102,7 +102,7 @@ Set the following parameters inside the corresponding php.ini file using your ow
 
 [source,php]
 ----
-post_max_size=16G
+post_max_size=10G
```
EParzefall, line 105:
```suggestion
post_max_size=10M
```

modules/admin_manual/pages/configuration/files/big_file_upload_configuration.adoc
```diff
@@ -127,7 +127,7 @@ You can configure php parameters also per virtual host - if you have access to t
 	DocumentRoot /var/www/owncloud
 	ServerName myowncloud.com
 
-	php_admin_value	post_max_size 16G
+	php_admin_value	post_max_size 10G
```
EParzefall, line 130:
```suggestion
	php_admin_value	post_max_size 10M
```

DeepDiver1975:
refs https://github.com/owncloud/security-tracker/issues/410

@EParzefall let me know if you have access to this .... most probably not .....

DeepDiver1975:
out latest approach is to stick with the php production defaults:
 - set `upload_max_filesize` to 2MB https://github.com/php/php-src/blob/master/php.ini-production#L865
 - set `post_max_size`  to 8MB https://github.com/php/php-src/blob/master/php.ini-production#L713

We are verifying this by QA next week -  afterwards next steps are following ....

EParzefall:
> refs https://github.com/owncloud/security-tracker/issues/410
> 
> @EParzefall let me know if you have access to this .... most probably not .....

@DeepDiver1975  Nope, no access.

TheRealBro:
> out latest approach is to stick with the php production defaults:
> 
>     * set `upload_max_filesize` to 2MB https://github.com/php/php-src/blob/master/php.ini-production#L865
> 
>     * set `post_max_size`  to 8MB https://github.com/php/php-src/blob/master/php.ini-production#L713
> 
> 
> We are verifying this by QA next week - afterwards next steps are following ....

For the docker image we use 8MB for both values, just for the record, because otherwise it would be a BREAKING change with changing ENV variables.

EParzefall:
Okay, I'll wait until you've reached a final decision after QA.

iasdeoupxe:
What about all the core things which e.g. seems to use `513M` for both:

https://github.com/search?q=repo%3Aowncloud%2Fcore%20post_max_size&type=code

DeepDiver1975:
We are still waiting for the verification by QA or whom ever ....

mmattel:
> We are still waiting for the verification by QA or whom ever ....

Adding @jnweiger @GeraldLeikam

jnweiger:
I don't have a good idea how to conduct manual tests here.
Is that needed? As long as it is just a documented recommodation, not much will break. 
But as soon as we update our docker containers with the lower limit, then any POST that tries something larger should break. 
I'd rely more on CI than on manual QA to find those...

DeepDiver1975:
> I'd rely more on CI than on manual QA to find those...

That would require integration tests on all apps - which we simply don't have. Let me think about options ....

DeepDiver1975:
> I don't have a good idea how to conduct manual tests here.
> Is that needed? As long as it is just a documented recommodation, not much will break.
> But as soon as we update our docker containers with the lower limit, then any POST that tries something larger should break.

doc change in this respect is as critical as docker image change - if too low functionalities will :boom:

TheRealBro:
> > I don't have a good idea how to conduct manual tests here.
> > Is that needed? As long as it is just a documented recommodation, not much will break.
> > But as soon as we update our docker containers with the lower limit, then any POST that tries something larger should break.
> 
> doc change in this respect is as critical as docker image change - if too low functionalities will 💥

Doc change is maybe even more critical than changing docker image, which we did, as Collabora runs in a separate container and is not affected by changes. Changing settings on the host, if running Collabora on the same host, might blow up things. But still, we should move on with testing and/or decision.

DeepDiver1975:
Just to be sure I explained it in detail enough - collabora and o365 (both using WOPI protocol on different api endpoints) are using POST to save the document back into owncloud.
As a result owncloud needs to accept POST request with a body of a few hundred MB.

Unfortunately there is no way to allow different post sizes on different endpoints .... at least I am not aware of any ....

PR #1160, opened 2023-11-02
Update manually-moving-data-folders.adoc

jnweiger:
No, apps don't have their own datadir. 
If we find an app that does that, it should be fixed to use the system datadirectory.

I also suggest to use a proper occ command to change the system datadirectory, instead of cumbersome manual editing instrutions.


mmattel (approved):
LGTM, thanks 👍 
Backport to 10.13 and 10.12

PR #1176, opened 2023-11-06
Update appliance firewall.adoc

jnweiger:
owncloud.org can be safely removed, we no longer use that domain.

176.9.114.147 has an invalid self singned certificate. It seem to be a hetzner machine. Without specifying a very good reason we should not ask anybody to trust this machine. I have no idea, by whom it is operated.

5.9.68.237 has a cert that identifies as "int.owncloud.com"; most likely it is also obsolete.


mmattel (approved):
LGTM, many thanks 👍 
Backport to 10.13 and 10.12

PR #1184, opened 2023-11-10
Update docs pertaining to the use of 'OWNCLOUD_TRUSTED_DOMAINS'

firephreek:
Provide additional information about the correct use and configuration of the trusted domains property when used in docker configurations.  Will hopefully reduce/resolve support requests relating to misuse of this value.

### DIFF
--- a/modules/admin_manual/pages/configuration/server/config_sample_php_parameters.adoc
+++ b/modules/admin_manual/pages/configuration/server/config_sample_php_parameters.adoc
@@ -58,7 +58,7 @@ passwords. This example is for documentation only, and you should never use it.
 ....
 
 === Define list of trusted domains that users can log into
-Specifying trusted domains prevents host header poisoning. This parameter reperesents a white list of approved IP addresses and hostnames that this server is known by.
+Specifying trusted domains prevents host header poisoning. This parameter reperesents a white list of approved IP addresses and hostnames that this server is known by / is used to access it. Wildcards, slash notation and ports are not supported.
 
 Do not remove this, as it performs necessary security checks.
 Please consider that for backend processes like background jobs or occ commands,
--- a/modules/admin_manual/pages/configuration/server/config_sample_php_parameters.adoc
+++ b/modules/admin_manual/pages/configuration/server/config_sample_php_parameters.adoc
@@ -58,7 +58,7 @@ passwords. This example is for documentation only, and you should never use it.
 ....
 
 === Define list of trusted domains that users can log into
-Specifying trusted domains prevents host header poisoning.
+Specifying trusted domains prevents host header poisoning. This parameter reperesents a white list of approved IP addresses and hostnames that this server is known by.
 
 Do not remove this, as it performs necessary security checks.
 Please consider that for backend processes like background jobs or occ commands,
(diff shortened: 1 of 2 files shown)


mmattel (commented):
Thanks for your PR.
You need to remove the changes of config.sample here and add them to the respective file in the core repo. There is an additional process I will take care of bringing these changes here to docs. Just reference this docs PR in the core one. If you are not doing this, the changes of config.sample here will get overwritten again.

firephreek:
config.sample?  I only have two changes in this commit: `config_sample_php_parameters.adoc and `index.adoc` and those changes are just to the docs themselves and the addition of one `Note:`.  Are you talking about the doc-strings in `core/config/config.sample.php`?  Happy to update those with the same notes here if that's what you mean.

mmattel:
You need to move the changes in `config_sample_php_parameters.adoc` --> https://github.com/owncloud/core/blob/master/config/config.sample.php

The source of the content of `config_sample_php_parameters.adoc` is in core which gets shipped as example config.sample with the product, though it looks we can change here. There is a process I run on a regular basis transporting any config.sample change into docs. Means stuff gets overwritten in docs if it is not in core. I will take care on that process when you do that PR in core.

iasdeoupxe:
Originating from:
- owncloud/core#41086
- owncloud-docker/server#432

modules/admin_manual/pages/configuration/server/config_sample_php_parameters.adoc
```diff
@@ -58,7 +58,7 @@ passwords. This example is for documentation only, and you should never use it.
 ....
 
 === Define list of trusted domains that users can log into
-Specifying trusted domains prevents host header poisoning.
+Specifying trusted domains prevents host header poisoning. This parameter reperesents a white list of approved IP addresses and hostnames that this server is known by.
```
iasdeoupxe, line 61:
```suggestion
Specifying trusted domains prevents host header poisoning. This parameter reperesents a white list of approved IP addresses and hostnames that this server is known by / is used to access it.
```

Maybe?

modules/admin_manual/pages/configuration/server/config_sample_php_parameters.adoc
```diff
@@ -74,6 +74,8 @@ the URL parameter in key `overwrite.cli.url` is used. For more details, please s
   ],
 ....
 
+NOTE: When defined via the `OWNCLOUD_TRUSTED_DOMAINS` property in docker, the values should be a comma delimited list with no white space. e.g. `OWNCLOUD_TRUSTED_DOMAINS=localhost,10.10.1.1`.  Wildcards and slash notation are not supported.
```
iasdeoupxe, line 77:
```suggestion
NOTE: When defined via the `OWNCLOUD_TRUSTED_DOMAINS` property in docker, the values should be a comma delimited list with no white space. e.g. `OWNCLOUD_TRUSTED_DOMAINS=localhost,10.10.1.1`.  Wildcards, slash notation and ports are not supported.
```

firephreek:
Pull request created.
https://github.com/firephreek/core/pull/1

PR #1208, opened 2023-12-04
Update manual_upgrade.adoc

morganbo85:
Suggest add -L attribute to chown command for setting permission so if users are using symlink for the data directory


mmattel:
Though I understand the intention, I do not agree with this change for the following reasons:

* We have an excellent [Script-Guided Installation](https://doc.owncloud.com/server/latest/admin_manual/installation/manual_installation/manual_installation.html#script-guided-installation) documentation which exactly covers this topic.
* A change here impacts much more in docs as we need to take care also on other locations. The easy setup uses hardcoded locations where one gets the freedom using links (preferred!) via the script. This script also covers upgrading. A clean seperation between the two is intentional.

Anyways, I welcome the effort taken and would like to encourage for more PR's if things get to your attention.

PR #1253, opened 2024-01-31
Update collabora_secure_view.adoc

pako81:
Remove the section about Secure View for public links as currently this feature is not implemented.


EParzefall (approved):

pako81:
should the Table of Contents also be updated?

EParzefall:
ToC should be updated automatically. Thanks for the fix!

PR #1290, opened 2024-03-25
fix: select encryption type is obsolete

IljaN:
Only master-key is supported now so the command is not available anymore.

Related Change: https://github.com/owncloud/encryption/pull/389

Backport to 10.14 and 10.13


mmattel:
@IljaN when was this removed, can you referencethe PR?
Note that this cange must be made in core and then transported via a process here.

IljaN:
@mmattel Edited top-post.

mmattel:
I have updated the content on locations where necessary.

mmattel (approved):

PR #1299, opened 2024-05-29
clarify sharing bits

jnweiger:
It is unclear, if only the listed numbers are supported. I'd think, that 1+2+4+8+16 are meant as bit patters that can be combined as needed. 15, 31 are probably just examples.

What does 'default for public link shares' mean? Is the 1 bit always added for these? I don't think so, as this would prevent building an anonymous file drop link.

See https://central.owncloud.org/t/share-files-via-api-without-delete-permissions/49350

### DIFF
--- a/modules/developer_manual/pages/core/apis/ocs-share-api.adoc
+++ b/modules/developer_manual/pages/core/apis/ocs-share-api.adoc
@@ -676,8 +676,11 @@ The permissions to set on the share.
 * 2 = update;
 * 4 = create;
 * 8 = delete;
-* 15 = read/write/delete;
 * 16 = share;
+
+Common example combinations are:
+
+* 15 = read/write/delete;
 * 31 = All permissions.
 
 | expireDate
--- a/modules/developer_manual/pages/core/apis/ocs-share-api.adoc
+++ b/modules/developer_manual/pages/core/apis/ocs-share-api.adoc
@@ -680,7 +680,7 @@ The permissions to set on the share.
 
 Common example combinations are:
 
-* 15 = read/write/delete;
+* 15 = read/write(update and create)/delete;
 * 31 = All permissions.
 
 | expireDate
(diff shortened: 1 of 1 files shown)


modules/developer_manual/pages/core/apis/ocs-share-api.adoc
```diff
@@ -676,8 +676,11 @@ The permissions to set on the share.
 * 2 = update;
 * 4 = create;
 * 8 = delete;
-* 15 = read/write;
 * 16 = share;
+
+Common example combinations are:
+
+* 15 = read/write/delete;
```
phil-davis, line 683:
```suggestion
* 15 = read/write(update and create)/delete;
```

In this case "write" is a word for "update+create". I don't know if that should be made explicit here, so that readers do not wonder where "write" came from.

phil-davis (approved):
LGTM

Someone else can decide if my suggestion is useful.

phil-davis:
> What does 'default for public link shares' mean?

It means that for an installation with the default config, if someone creates a public link share without specifying permissions, then the public link share will have just read access.
"
The user (their client app) can request different permissions than the default, to get an "anonymous file drop" public link share etc.

phil-davis (approved):

modules/developer_manual/pages/core/apis/ocs-share-api.adoc
```diff
@@ -676,8 +676,11 @@ The permissions to set on the share.
 * 2 = update;
 * 4 = create;
 * 8 = delete;
-* 15 = read/write;
 * 16 = share;
+
+Common example combinations are:
+
+* 15 = read/write(update and create)/delete;
```
mmattel, line 683: Hmmm, maybe I got it wrong, but in my calculation:
```
 2 ... update
 4 ... create
 8 ... delete
14 and not 15
```
Additionally `31` (see below) is also not the sum of the upper ones if I am correct with my view...

@jnweiger can you bring some light into it?

phil-davis: read = 1, 1+2+4+8=15

another 16 for share gets to 31, the maximum

mmattel: I missed the read = 1 🤦‍♂️

mmattel:
Backport to 10.14 and 10.13

mmattel (approved):

PR #1302, opened 2024-06-05
Fix command typo in generate-index.js

phil-davis:
`console.warning(err)` --> `console.warn(err)` :man_facepalming:

Backport to 10.14


phil-davis:
This is a "copy" of #1301
I was hoping that this would trigger drone to start, but it did not.
I made a completely different PR #1302 and drone does not start there either.
"something is wrong"

phil-davis:
There was a problem with repo settings, which is fixed now. So closing this.

PR #1327, opened 2024-06-25
doc: review migrating to kiteworks

phil-davis:
This has been touched a bit recently, so just cleaning up some small wording issues.

PR #1332, opened 2024-06-25
Revert "doc: review migrating to kiteworks"

mmattel:
Reverts owncloud/docs-server#1327

There was a conflict issue where old text sneaked in that was formerly removed.
The backports need to be deleted too and the language fixes need to be redone post merging.

@phil-davis Sorry for the inconvinience...


phil-davis (commented):
I don't know why you want to revert this. The changes made to master were all good things.

The problem was only in the back-ports. When back-porting, conflicts need to be correctly resolved.

mmattel:
> The problem was only in the back-ports. When back-porting, conflicts need to be correctly resolved.

I think it was not my day and assumed that the issue origined in the master merge whithout taking a closer look. BP never had such a problem before...

You can close that PR and delete the branch if all is fine. Back in duty on the 1st 😄 👋

PR #1342, opened 2024-06-26
clarify token_auth_enforced in oauth2.adoc

jnweiger:
The effect of token_auth_enforced is officially described in the sample config php. There it is only discussed in the context of basic auth vs. app tokens. 

Admins migrating from basic auth to oauth2 may eventually come here, and wonder what exact effect token_auth_enforced has. oauth2 either way, true or false, but true does not do a hard enforce.


jnweiger:
Side note: Server UI speaks about: app token, auth token, token, app password, app passcode.
Can we clarify in the docs, that all this is the same?

mmattel:
> Can we clarify in the docs, that all this is the same?

We could 🤣

mmattel (approved):

mmattel:
Backport to 10.15 and 10.14

PR #1361, opened 2024-07-30
Change php-dev version used for Kerberos

d7oc:
The current installation description will install all `php-dev` packages. So on newer systems this will install a complete php 8 stack as well even if there was only php 7.4 installed so far. So the description should be change to install the dev package of the appropriate version.


mmattel (approved):
Backport to 10.14 and 10.15

PR #1378, opened 2024-09-29
Update caching_configuration.adoc

scottbomb:
The code "memcache.local" does nothing and I still get the warning about using db caching. With this set to "memcache.locking" like you show for similar other configs, that is the one that worked.

### DIFF
--- a/modules/admin_manual/pages/configuration/server/caching_configuration.adoc
+++ b/modules/admin_manual/pages/configuration/server/caching_configuration.adoc
@@ -203,7 +203,7 @@ To use APCu, add this line to `config.php`:
 
 [source,php]
 ----
-'memcache.local' => '\OC\Memcache\APCu',
+'memcache.locking' => '\OC\Memcache\APCu',
 ----
 
 With that done, refresh your ownCloud admin page, and the cache warning should disappear.


iasdeoupxe:
This doesn't look right to me, page below https://doc.owncloud.com/server/10.15/admin_manual/configuration/server/caching_configuration.html#configuration-recommendations-based-on-type-of-deployment shows various examples like e.g.:

```
'filelocking.enabled' => true,
'memcache.local' => '\OC\Memcache\APCu',
'memcache.locking' => '\OC\Memcache\Redis',
'redis' => [
    'host' => 'localhost',
    'port' => 6379,
],
```

mmattel:
There are 3 different memcache "types" which are `local`, `locking` and `distributed`.
We should not mix up file locking and caching, these are totally different things.

**Quotes:**

```
APCu is an in-memory key-value store for PHP. Keys are of type string and values
can be any PHP variables. APCu only supports userland caching of variables.
```
```
Redis is an open source in-memory data store that can be used as a database,
cache, or message broker.
```
```
Redis Lock, also known as Distributed Locking with Redis, is a technique used to
coordinate access to shared resources in a distributed environment.
```

Reading this makes the usecase very much clear and docs are right.

If you experience problems using/defining the cache/lock, file an issue in https://github.com/owncloud/core

Personal note, I use redis for both, chaching and locking and have no issues:

```
  'filelocking.enabled' => 'true',
  'memcache.local' => '\\OC\\Memcache\\Redis',
  'memcache.locking' => '\\OC\\Memcache\\Redis',
  'memcache.distributed' => '\\OC\\Memcache\\Redis',
  'redis' =>
  array (
    'host' => '/var/run/redis/redis-server.sock',
    'port' => 0,
    'timeout' => 0,
    'dbindex' => 0,
  ),
```

Closing this for now as I dont see a docs issue here.

PR #1404, opened 2024-12-04
chore: drop QNAP from documentation

DeepDiver1975:


mmattel (approved):
LGTM 👍 
grep does not return any xrefs containing 'qnap'

Backport to 10.14 and 10.15

PR #1428, opened 2025-01-21
[10.14] [1426] Extension: Implement dynamic loading attributes

mmattel:
Backport of #1426


phil-davis (commented):
`yarn.lock` is not deleted.
Why is that?

mmattel:
> `yarn.lock` is not deleted. Why is that?

good question...
I deleted it, and pushed the commit.
should be gone now

phil-davis (approved):

PR #1511, opened 2026-04-09
feat: oc11 to support docker only deployments

DeepDiver1975:


modules/admin_manual/partials/maintenance/major_release_note.adoc
```diff
@@ -1,6 +1,6 @@
 [IMPORTANT]
 ====
-When upgrading, also check the minimum and maximum supported PHP version of the ownCloud target release. An ownCloud release may require a particular minimum and/or maximum PHP version. Check that the PHP version provided by the Operating System meets the requirements. For details see the https://doc.owncloud.com/docs_main/next/server_release_notes.html[Server Release Notes] and the xref:{latest-server-version}@server:admin_manual:installation/system_requirements.adoc[System Requirements] for the latest Release.
+When upgrading, also check the minimum and maximum supported PHP version of the ownCloud target release. An ownCloud release may require a particular minimum and/or maximum PHP version. Check that the PHP version provided by the Operating System meets the requirements. For details see the https://doc.owncloud.com/docs_main/next/server_release_notes.html[Server Release Notes] and the xref:server:admin_manual:installation/system_requirements.adoc[System Requirements] for the latest Release.
```
DeepDiver1975: This is non-sense. The provided docker image will ship the correct php version and modules....

DeepDiver1975:
## PR Review — feat: oc11 to support docker-only deployments

**Status:** DRAFT | +2248 / -1733

---

### Overview

This PR migrates all ownCloud Server documentation to assume a Docker-only deployment model. It:
- Renames/replaces the `occ-command-example-prefix` attribute with `occ-command-example-prefix-docker` (resolved as `docker compose exec owncloud occ`)
- Rewrites the `occ_command.adoc` intro section to focus entirely on Docker
- Updates `docker-compose.yml` with several structural improvements (named networks, variable-driven image tags, long-form volume syntax)
- Deletes the command-line installation commands doc
- Adds a `NOTE: Run this command inside the ownCloud Docker container.` after every single occ code block

---

### Issues & Concerns

#### Security — High

- **Exposed database and cache ports in `docker-compose.yml`:** Ports `3306` and `6379` are now published to the host. By default Docker binds these to `0.0.0.0`, making MariaDB and Redis reachable from outside the host. For a documentation/example compose file that new users copy-paste, this is a serious security footgun. These services communicate only within `oc_network` and do not need host-level port bindings.

---

#### Breaking Changes — Should Be Documented

- **Volume renamed from `files` to `oc_files`:** Any existing user following the prior guide and then pulling this update will have orphaned data in the old `files` volume. A migration note or upgrade warning is needed.

- **Env var `OWNCLOUD_VERSION` → `OWNCLOUD_IMAGE`:** Existing `.env` files will silently fail because the variable name changed. Either keep backward compatibility or explicitly call out this rename in an upgrade notice.

- **Page aliases removed from `occ_command.adoc`:** The large `:page-aliases:` block (covering ~30 paths) was deleted entirely. These aliases exist specifically to keep old bookmarks and cross-references alive. Removing them will break every inbound link from the published site and other docs repos that reference those paths.

---

#### Documentation Quality

- **Repetitive notes:** The phrase `NOTE: Run this command inside the ownCloud Docker container.` appears after *every single* code block — more than 100 times across the diff. This creates significant noise. A single note at the top of the `occ_command.adoc` "Running occ" section (and possibly a short admonition box at the top of frequently-used sub-pages) would communicate the same information far more cleanly. Repeating it per-command suggests the docs were mechanically transformed rather than deliberately revised.

- **`[source,docker]` is non-standard:** AsciiDoc/Antora does not recognise `docker` as a syntax-highlighting language identifier. Use `[source,bash,subs="attributes+"]` (as was done previously) or `[source,console,subs="attributes+"]` to get consistent rendering.

- **Version mismatch in examples:** The example `occ status` output now shows `11.0.0`, but `latest-server-version` in `global-attributes.yml` is still `10.15`. Either update both or keep the examples at the current release version to avoid confusing readers.

---

#### Structural / Correctness

- **Healthcheck removed from ownCloud service:** The previous compose file used the ownCloud healthcheck as a `depends_on` condition target. Removing it without checking whether anything downstream still references it could silently break container orchestration ordering. Verify whether `depends_on: condition: service_healthy` is still used anywhere in the compose or docs for this service.

- **Inconsistent volume definitions:** `oc_files` uses the detailed long-form syntax with a bind-mount template comment, but `mysql` and `redis` still use the terse short form. Apply a consistent style across all three.

- **Deleted `_command_line_installation_commands.adoc`:** The `maintenance:install` command is useful for advanced scenarios (CI/CD, bare-metal fallback). Removing the doc entirely (rather than moving it to a "legacy" or "advanced" section) discards that content. Check whether the alias pointing to it is also gone from `occ_command.adoc` — it was in the page-aliases block that was deleted.

- **SELinux references silently removed:** `configuration.adoc` and `harden_server.adoc` had links to `selinux_configuration.adoc` that were stripped without explanation. If Docker removes the SELinux requirement, that should be stated explicitly; if the docs still apply, the links should stay.

---

### Summary

The direction is correct for a Docker-first docs strategy, but the PR has three items that need to be resolved before merging:

1. Remove the exposed `3306`/`6379` host port bindings — they are a security risk for copy-paste users.
2. Restore the page aliases in `occ_command.adoc` or explicitly redirect them — removing them silently breaks published URLs.
3. Reduce the repetitive Docker container notes to a single per-page or per-section admonition.

The version mismatch (`11.0.0` vs `10.15`) and the breaking env-var / volume renames also need either updates or explicit migration guidance before this is ready to merge.

🤖 *Review generated with [Claude Code](https://claude.com/claude-code)*

modules/admin_manual/pages/configuration/database/db_conversion.adoc
```diff
@@ -59,7 +59,7 @@ After you have restarted the database, run the following occ command in your own
 
 [source,bash,subs="attributes+"]
 ----
-{occ-command-example-prefix} db:convert-type [options] type username hostname database
```
DeepDiver1975, line 62: db:convert-type was dropped with oc11

modules/developer_manual/pages/testing/drone.adoc
```diff
@@ -21,7 +21,7 @@ This document will show you how to run drone tests locally.
 +
 [source,console,subs="attributes+"]
 ----
-{occ-command-example-prefix} maintenance:install
+{occ-command-example-prefix-docker} maintenance:install
```
DeepDiver1975: drone is dead - remove this page

mmattel: removed

mmattel:
If we need it again to find and replace multi line patterns (Claude issue):
`find . -type f -name '*.adoc' -exec sed -i -z -E 's|\nNOTE: Run this command inside the ownCloud Docker container.\n||gm' '{}' \;`

modules/admin_manual/pages/configuration/database/db_conversion.adoc
mmattel: removed

modules/admin_manual/pages/configuration/files/previews_configuration.adoc
```diff
@@ -84,35 +94,29 @@ and change:
 rights="none" --> rights="read|write"
 ----
 
-After changing the policy file for ImageMagick, restart your Apache web server or your php-fpm service.
-ImageMagick uses ghostscript for converting PDF. Please also make sure, that ghostscript is installed and the PDF format is listed when invoking the following command:
+After changing the policy file for ImageMagick, restart your container. Check that the PDF format is listed when invoking the following command:
```
DeepDiver1975, line 99: We need to find a different approach. Docker container and images need to be treated as immutable.

As already explained - maybe not detailed enough - never change anything within the image and container. This also includes the ownCloud config file. ownCloud is to be configured only via env vars. 

In case anything is missing in the container as of now: treat this as a bug in the image and we work on a solution either by extending the emv vars of declaring a feature as not accessible anymore (e.g. hardening of SVG - we will not ship exploitable images)

modules/admin_manual/pages/configuration/server/harden_server.adoc
```diff
@@ -71,7 +58,7 @@ With this information, you can begin customizing a rate-limiting solution specif
 
 == Operating System
 
-=== Enable hardening modules such as SELinux
+=== Enable Hardening Modules Such as SELinux
```
DeepDiver1975: Drop selinix. We don't have it in the image and will not add it. Users cannot enable it. Pointless....

mmattel: done

modules/admin_manual/pages/configuration/files/previews_configuration.adoc
mmattel: We keep it as discussed for the time being

mmattel:
Because this PR is already with many changes, we merge now and add missing stuff in another PR.

mmattel (approved):

PR #1577, opened 2026-07-30
docs: require explicit docker image versions

DeepDiver1975:
Every documented `owncloud/server` reference on this branch now carries an
explicit version tag, and `latest` is no longer presented as a valid option. Same
change as owncloud/docs-server#1576 (10.16) and #1575 (master).

`owncloud/server:latest` does not track the release line documented here — it
follows whatever the newest published image happens to be — so a reader copying
the examples, or a container restart, can end up on a different version than the
documentation describes. There are also no rolling major or minor tags
(`owncloud/server:10.15` and `:10` both 404 on Docker Hub), so a full version is
the only correct thing to name.

This applies to `owncloud/server` only. An earlier revision of this PR also
pinned the selenium and inbucket images in the UI testing guide; those pins are
reverted, since they fix no documented-version mismatch and nothing in this repo
keeps them current.

## Changes

`modules/admin_manual/pages/installation/docker/index.adoc`

* Adds a `docker-image-version` page attribute set to the literal `10.15.3`, the
  top of this release line. This intentionally does *not* reuse
  `{latest-server-version}`: that is a global site attribute tracking the latest
  release of the product, and the value that wins is the aggregator's
  (`owncloud/docs` `global-attributes.yml`), not this branch's `antora.yml` — which
  is why the published page currently renders `OWNCLOUD_VERSION=10.16` here.
* Tags the quick-evaluation `docker run` (`owncloud/server:10.15.3`); it was
  untagged, which resolves to `latest`.
* `OWNCLOUD_VERSION` in the `.env` example now names the explicit version, and the
  required-settings table no longer gives `latest` as its example value.
* Adds a note explaining why `OWNCLOUD_VERSION` must be pinned.

`modules/admin_manual/pages/useful_pages.adoc`

* Fixes the xrefs into the *latest* server version, which the page deliberately
  targets but addressed with this branch's page paths. The 11.0 release is
  docker-only and reorganized accordingly, so three of the five links break as
  soon as `latest-server-version` advances.

## Testing

`npm run antora-local` (Node 22, as pinned by this branch's `ci.yml`) builds with
no errors and no warnings. The rendered page shows `OWNCLOUD_VERSION=10.15.3`,
`owncloud/server:10.15.3`, and no occurrence of `latest` in the settings table.
The `10.15.3` tag was verified present on Docker Hub.

## Note on timing

10.15 is about to be dropped from the published site and archived as
`x_archived_10.15`, so this content will not be published for long. It is worth
merging anyway: it keeps the branch correct for the window before the
`owncloud/docs` change lands, and should the branch ever be un-archived it is not
left documenting an image tag that points somewhere else. This PR needs to merge
**before** the branch is renamed.

### DIFF
--- a/modules/admin_manual/pages/installation/docker/index.adoc
+++ b/modules/admin_manual/pages/installation/docker/index.adoc
@@ -5,6 +5,10 @@
 :linux-server-doc-url: https://docs.linuxserver.io/faq
 :docker-compose-build-url: https://docs.docker.com/compose/compose-file/build/
 :description: ownCloud can be installed using the official ownCloud Docker image.
+// the image version documented on this branch. this is intentionally a literal
+// and not an attribute, because the global -version attributes track the latest
+// release of the product and not this branch.
+:docker-image-version: 10.15.3
 
 == Introduction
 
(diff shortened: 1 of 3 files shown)


phil-davis (dismissed):

DeepDiver1975:
Added a second commit that fixes `useful_pages.adoc`, matching owncloud/docs-server#1576.

That page deliberately xrefs into the *latest* server version rather than this branch, but used this branch's page paths. 11.0 is a docker-only release and reorganized accordingly, so three of the five links break once `latest-server-version` advances: `installation/index.adoc` no longer exists, the docker page moved to `installation/installing_with_docker.adoc`, and `maintenance/manual_upgrade.adoc` moved under `maintenance/upgrading/`.

Unlike on 10.16 this cannot fail the site build, since this branch is being dropped from the published site — but it is the same belt-and-braces reasoning as the `:latest` fix: keep the branch content correct in case it is ever un-archived.

`npm run antora-local` on this branch: 0 errors, 0 warnings.

phil-davis (dismissed):

jvillafanez (approved):

kw-tmueller (approved):

